PR #2, opened 2019-02-17
Replace core code with testHook from react-testing-library

### LINKED ISSUE #1: Merge with testHook - move package here?
Hi,
Some folks that worked on hook test utils in `react-testing-library` (@kentcdodds, @donavan) are thinking about extracting `testhook` into its own package. The API is pretty similar to what you have here - would you consider reviewing the current state of that API and merging it with `react-hooks-testing-library`?

Discussion: https://github.com/kentcdodds/react-testing-library/issues/302

WIP Docs for testHook: https://github.com/alexkrolick/testing-library-docs/pull/32

Next steps:

- [x] Review and port the API if it is an improvement
- [ ] Add a docs page in the "ecosystem" section of the [docs repo][docsgh]
- [ ] Port examples in the docs [PR](https://github.com/alexkrolick/testing-library-docs/pull/32) to extracted library

[docsgh]: https://github.com/alexkrolick/testing-library-docs
[docssite]: https://testing.library.com
[docspr]: https://github.com/alexkrolick/testing-library-docs/pull/32

mpeyper:
So my review found that `useHook` and `testHook` offered practically identical features.  The `testHook` implementation is simpler than what I had in `useHook` so I took that as the base for moving forward.

I made one change to `testHook` to allow the callback to take props (similar to`useHook`).  In my opinion, this offers a slightly nicer story for updating the values passed to the hook to test how the result changed when inputs changed.  It's also backwards compatible with the old `testHook` so whatever people were doing before will continue to work when migrating across.

I'm also not a big fan of the `options.wrapper` usage.  Having to wrap a context provider in order to pass the `value` to it seemed clunky in practice.  Even the tests in `react-testing-library` had to [turn off lint warnings](https://github.com/kentcdodds/react-testing-library/blob/master/src/__tests__/test-hook.js#L53) when using it.  I'll sleep on it and see if I can't come up with something a little bit more ergonomic for it but I'll probably keep `options.wrapper` regardless for backwards compatibility reasons, at least to begin with.

Other than that, I'm pretty happy with this.

Resolves #1

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,28 +64,28 @@ const useTheme = (initialTheme) => {
 }
 
 // useTheme.test.js
-import { useHook, cleanup } from 'react-hooks-testing-library'
+import { testHook, cleanup, act } from 'react-hooks-testing-library'
 
 describe('custom hook tests', () => {
   afterEach(cleanup)
 
   test('should use theme', () => {
-    const { getCurrentValue } = useHook(() => useTheme('light'))
+    const { result } = testHook(() => useTheme('light'))
 
-    const theme = getCurrentValue()
+    const theme = result.current
 
     expect(theme.primaryLight).toBe('#FFFFFF')
     expect(theme.primaryDark).toBe('#000000')
   })
 
   test('should update theme', () => {
-    const { getCurrentValue, act } = useHook(() => useTheme('light'))
+    const { result } = testHook(() => useTheme('light'))
 
-    const { toggleTheme } = getCurrentValue()
+    const { toggleTheme } = result.current
 
     act(() => toggleTheme())
 
-    const theme = getCurrentValue()
+    const theme = result.current
 
     expect(theme.primaryLight).toBe('#000000')
     expect(theme.primaryDark).toBe('#FFFFFF')
@@ -97,11 +97,13 @@ describe('custom hook tests', () => {
       dark: { primaryLight: '#CCBBAA', primaryDark: '#AABBCC' }
     }
 
-    const { getCurrentValue, addContextProvider } = useHook(() => useTheme('light'))
+    const wrapper = ({ children }) => (
+      <ThemesContext.Provider value={customThemes}>{children}</ThemesContext.Provider>
+    )
 
-    addContextProvider(ThemesContext, { value: customThemes })
+    const { result } = testHook(() => useTheme('light'), { wrapper })
 
-    const theme = getCurrentValue()
+    const theme = result.current
 
     expect(theme.primaryLight).toBe('#AABBCC')
     expect(theme.primaryDark).toBe('#CCBBAA')
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "react-hooks-testing-library",
-  "version": "0.2.4",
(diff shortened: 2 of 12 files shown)


alexkrolick:
Looks good to me 👍

- Would you mind pulling in the unit tests from RTL for testHook? 
- Can you show the usage of the callback argument you mentioned?

mpeyper:
Dropped those tests in and they worked without any changes (other than formatting to match my prettier config).

For changing the props, I added a new `option` called `initialProps` that gets passed into the callback which can be changed when calling `rerender`.

e.g. given the hook

```js
const sideEffect = { '1': false, '2': false }

function useSideEffect(id) {
  useEffect(() => {
    sideEffect[id] = true
    return () => {
      sideEffect[id] = false
    }
  }, [id])
}
```

You can change the `id` in your test like so

```js
test('should handle useEffect hook', () => {
  const { rerender, unmount } = testHook(({ id }) => useSideEffect(id)), { initialProps: { id: 1 } })

  expect(sideEffect[1]).toBe(true)
  expect(sideEffect[2]).toBe(false)

  rerender({ id: 2 })

  expect(sideEffect[1]).toBe(false)
  expect(sideEffect[2]).toBe(true)

  unmount()

  expect(sideEffect[1]).toBe(false)
  expect(sideEffect[2]).toBe(false)
})
```

Maximilianos:
Hey @mpeyper thanks for putting this together, I started using this branch in a project of mine and all is 👍 at the moment. 

Minor naming comment: what do you think of calling the method `renderHook` (or even `render`) instead of `testHook` to be closer to the `render` method of `react-testing-library` and be closer to the `rerender` method that is returned from `testHook` currently?

mpeyper:
This branch actually renamed it from `useHook` which was a fun riff on the hook naming convention.

I also find `testHook` to feel a bit off, but the intention here was for 1-1 compatibility with react-testing-library.

I would consider an alias export though as I do like your suggestion.

Maximilianos:
Yeah, I saw that it was `useHook` before. Great riff 😄! but also suffers from the same issues with `testHook` regarding the differences in naming and `rerender`. Given the `testHook` feature in RTL was never really publicized I assume it was kind of in beta, so the name could still have changed?

Maybe @alexkrolick or @donavon have stronger opinions on this?

I anyway now do something like:
```javascript
function renderHook(initialProps = {}) {
  return testHook(({ query }) => useMovieSearch(query), { initialProps })
}
``` 
so, it's up to you 😄

mpeyper:
with `useHook`, there wasn't the naming conflict because it would just rerender on each `getCurrentValue()` calls.  I prefer the new sematics, but agree that `renderHook` feels better.

I'm going to change the export to be `renderHook` and create a `testHook` that looks something like

```
function testHook(...params) {
  console.warn('testHook is deprecated.  Please use renderHook instead.')
  return renderHook(...params)
}

mpeyper:
Ok, I'm going to merge and publish this v0.3.0.

I'm still not a fan of the `wrapper` API, but nothing I do will be backwards compatible, so I'll leave it as is for now and tackle it another day.  I'll open an RFC for it and see if anyone else has ideas for it.

PR #8, opened 2019-02-28
Add support for TypeScript

### LINKED ISSUE #7: TypeScript Typings
React Testing Library ships with TypeScript types, and when moving to use `renderHook` from this react-hooks-testing-library, there are no types declared.

This seems like the types could mostly be re-exported from react-testing-library, and pull in what it had for `testHook` before removing it in v6.0.0: https://github.com/kentcdodds/react-testing-library/commit/12d934abd14990d0164c83f1574b13f6d386ce3f#diff-494415c3af35245ecae73f5d42d8b331L55

otofu-square:
Close #7 

Introduce TypeScript type definitions :tada:

- [x] Add some packages to devDependencies(`typescript`, `typings-tester`, `@types/react`)
- [x] Add `index.d.ts` to define typings
- [x] Add test for `index.d.ts` with [typings-tester](https://github.com/aikoven/typings-tester)
  - Inspired by the typing tests in [Redux repo](https://github.com/reduxjs/redux/blob/master/test/typescript.spec.js)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,13 +1,13 @@
 import { cleanup, act, RenderOptions, RenderResult } from 'react-testing-library'
 
-export function renderHook<P extends any, T extends (...args: [P]) => any>(
-  callback: (_: P) => ReturnType<T>,
+export function renderHook<P extends any, R extends any>(
+  callback: (...args: [P]) => R,
   options?: {
     initialProps?: P
   } & RenderOptions
 ): {
   readonly result: {
-    current: ReturnType<T>
+    current: R
   }
   readonly unmount: RenderResult['unmount']
   readonly rerender: (hookProps?: P) => void
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,6 @@
 import { cleanup, act, RenderOptions, RenderResult } from 'react-testing-library'
 
-export function renderHook<P extends any, R extends any>(
+export function renderHook<P, R>(
   callback: (...args: [P]) => R,
   options?: {
     initialProps?: P
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,9 +1,9 @@
 import { cleanup, act, RenderOptions, RenderResult } from 'react-testing-library'
 
-export function renderHook<T extends (...args: any[]) => any>(
-  callback: T,
+export function renderHook<P extends any, T extends (...args: [P]) => any>(
+  callback: (_: P) => ReturnType<T>,
   options?: {
-    initialProps?: Parameters<T>[0]
+    initialProps?: P
     options?: RenderOptions
   }
 ): {
@@ -11,7 +11,7 @@ export function renderHook<T extends (...args: any[]) => any>(
     current: ReturnType<T>
   }
   readonly unmount: RenderResult['unmount']
-  readonly rerender: (hookProps?: Parameters<T>[0]) => void
+  readonly rerender: (hookProps?: P) => void
 }
 
 export const testHook: typeof renderHook
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,7 +1,7 @@
 import { cleanup, act, RenderOptions, RenderResult } from 'react-testing-library'
 
 export function renderHook<P, R>(
-  callback: (...args: [P]) => R,
+  callback: (props: P) => R,
   options?: {
     initialProps?: P
   } & RenderOptions
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,19 @@
+import { cleanup, act, RenderOptions, RenderResult } from 'react-testing-library'
+
+export function renderHook<T extends (...args: any[]) => any>(
+  callback: T,
+  options?: {
+    initialProps?: Parameters<T>[0]
+    options?: RenderOptions
+  }
+): {
+  readonly result: {
+    current: ReturnType<T>
+  }
+  readonly unmount: RenderResult['unmount']
+  readonly rerender: (hookProps?: Parameters<T>[0]) => void
+}
+
+export const testHook: typeof renderHook
+
+export { cleanup, act }
--- a/index.d.ts
+++ b/index.d.ts
@@ -4,8 +4,7 @@ export function renderHook<P extends any, T extends (...args: [P]) => any>(
   callback: (_: P) => ReturnType<T>,
   options?: {
     initialProps?: P
-    options?: RenderOptions
-  }
+  } & RenderOptions
 ): {
   readonly result: {
     current: ReturnType<T>
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "@babel/plugin-transform-modules-commonjs": "^7.2.0",
     "@babel/preset-env": "^7.3.4",
     "@babel/preset-react": "^7.0.0",
+    "@types/react": "^16.8.5",
     "all-contributors-cli": "^6.1.2",
     "babel-eslint": "^10.0.1",
     "babel-plugin-module-resolver": "^3.2.0",
@@ -39,7 +40,9 @@
     "prettier-eslint": "^8.8.2",
     "prettier-eslint-cli": "^4.7.1",
     "react": "^16.8.3",
-    "react-dom": "^16.8.3"
+    "react-dom": "^16.8.3",
+    "typescript": "^3.3.3333",
+    "typings-tester": "^0.3.2"
   },
   "peerDependencies": {
     "react": "^16.8.0",
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "0.3.4",
   "description": "Simple component wrapper for testing React hooks",
   "main": "lib/index.js",
+  "typings": "./index.d.ts",
   "author": "Michael Peyper",
   "repository": {
     "type": "git",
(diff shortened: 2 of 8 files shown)


test/typescript/renderHook.ts
```diff
@@ -0,0 +1,60 @@
...
+type InitialTheme = typeof DARK | typeof LIGHT | undefined
+
+const themes = {
+  light: { primaryLight: '#FFFFFF', primaryDark: '#000000' },
+  dark: { primaryLight: '#000000', primaryDark: '#FFFFFF' }
+}
+
+const ThemesContext = createContext(themes)
+
+const useTheme = (initialTheme: InitialTheme = DARK) => {
```
otofu-square: Add minimum typing tests with the sample code in README.md.

otofu-square:
@mpeyper 
Please review this PR and let me know if you have any questions. thx!

otofu-square:
Created minimum TS project with react-hooks-testing-library: https://github.com/otofu-square/react-hooks-testing-library-sandbox

test/typescript/renderHook.ts
mpeyper: I'm not a big fan of that example, but I wanted something that used a few different hooks, so it became a little contrived.  Happy for you to come up with something better if you want, or just keep it as is that's easier.

index.d.ts
```diff
@@ -0,0 +1,18 @@
...
+  } & RenderOptions
+): {
+  readonly result: {
+    current: ReturnType<T>
+  }
+  readonly unmount: RenderResult['unmount']
+  readonly rerender: (hookProps?: P) => void
+}
+
+export const testHook: typeof renderHook
```
mpeyper: nice

mpeyper:
I'm reviewing this, but I'm far from a Typescript expert (I don't use it day to day, just created basic type definitions for some of my other libraries before, and not of them are as cool as what you've done here), so I'm sorry for any dumb questions.  I'd also like it if someone that does use Typescript regularly could chip in and give me an indiation of whether the types here make sense.

index.d.ts
```diff
@@ -0,0 +1,18 @@
+import { cleanup, act, RenderOptions, RenderResult } from 'react-testing-library'
+
+export function renderHook<P extends any, T extends (...args: [P]) => any>(
+  callback: (_: P) => ReturnType<T>,
```
mpeyper: Where does `ReturnType` come from?

mpeyper: Nevermind, I found [something about it in the TS docs](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html)

mpeyper:
This looks great to me.  Waaaaaay better than anything I could have done myself.  As I said before, I'd love a TS user to put a comment on here with approval before merging.

Also can you please add yourself as a contributor to the README ([by running this](https://github.com/mpeyper/react-hooks-testing-library/blob/master/package.json#L17)).  I haven't got around to setting up a contributing document yet, so I don't blame you for not knowing this.

If anyone TS users do review this, please submit a PR adding yourself as a contributer as well.

test/typescript/renderHook.ts
otofu-square: > so it became a little contrived

I see.

> Happy for you to come up with something better

I'll give it a try soon 👍

otofu-square:
> Also can you please add yourself as a contributor to the README

Oh, sorry! Will work on it soon.

otofu-square:
> If anyone TS users do review this

Let me ask someone I know.

test/typescript/renderHook.ts
otofu-square: Replace `useTheme` hook with `useCounter` here 3cb8f92.
Also fix type definitions because I just realized they are broken.

index.d.ts
```diff
@@ -0,0 +1,18 @@
...
+  } & RenderOptions
+): {
+  readonly result: {
+    current: R
+  }
+  readonly unmount: RenderResult['unmount']
+  readonly rerender: (hookProps?: P) => void
+}
+
+export const testHook: typeof renderHook
```
tatsushitoji: It'll be unnecessary someday, but it 's so smart 👍

index.d.ts
```diff
@@ -0,0 +1,18 @@
+import { cleanup, act, RenderOptions, RenderResult } from 'react-testing-library'
+
+export function renderHook<P extends any, R extends any>(
```
mpeyper: what is the difference between `renderHook<P extends any, R extends any>` and `renderHook<P, R>`?

index.d.ts
```diff
@@ -0,0 +1,18 @@
...
+  callback: (...args: [P]) => R,
```
mpeyper: is the `...args: [P]` saying that there can be more than 1 arg, or is it some syntax to describe the first (and only) arg?

mpeyper:
How do the typings look when the callback returns `void` (i.e. a hook that does not actually return anything)? Is it worth having a test around that case?  I'm not sure how often that will actually occur in the wild so I'm not too concerned, more curious than anything.

index.d.ts
otofu-square: The latter is true.
We can use *Tuple type* to extract args types.

```js
function foo<A1, A2, A3>(...args: [A1, A2, A3]): { a1: A1, a2: A2, a3: A3 } {
    return { a1: args[0], a2: args[1], a3: args[2] };
}

foo(1, '', true); // => the return value type is `{ a1: number, a2: string, a3: boolean }`
```

https://www.typescriptlang.org/play/index.html#src=function%20foo%3CA1%2C%20A2%2C%20A3%3E(...args%3A%20%5BA1%2C%20A2%2C%20A3%5D)%3A%20%7B%20a1%3A%20A1%2C%20a2%3A%20A2%2C%20a3%3A%20A3%20%7D%20%7B%0D%0A%20%20%20%20return%20%7B%20a1%3A%20args%5B0%5D%2C%20a2%3A%20args%5B1%5D%2C%20a3%3A%20args%5B2%5D%20%7D%3B%0D%0A%7D%0D%0A%0D%0Aconst%20res%3A%20%7B%20a1%3A%20number%2C%20a2%3A%20string%2C%20a3%3A%20boolean%20%7D%20%3D%20foo(1%2C%20''%2C%20true)%3B

otofu-square: I used `[P]` to extract first argment type here.

index.d.ts
otofu-square: `rednerHook<P, R>` is the same as `renderHook<P extends Object, R extends Object>`.
Maybe it's more simple to use `<P, R>` instead of `<P extends any, R extends any>` since `P` and `R` will be inferred by giving `callback` & `initialProps`.

otofu-square:
> I'm not sure how often that will actually occur in the wild so I'm not too concerned, more curious than anything.

IMHO, I think `result` which returns from `renderHook` is unnecessary if hook returns `void`.

> Is it worth having a test around that case?

Maybe it's worth to test that `result.current` has `void` type when hook returns `void` 🤔

index.d.ts
FredyC: If you want just a first arg, shouldn't it work like this as well? I am wondering what's the difference?
```js
  callback: (props: P) => R,
```

index.d.ts
```diff
@@ -0,0 +1,18 @@
...
+  options?: {
+    initialProps?: P
+  } & RenderOptions
+): {
+  readonly result: {
+    current: R
+  }
+  readonly unmount: RenderResult['unmount']
+  readonly rerender: (hookProps?: P) => void
```
FredyC: Feels unnecessary to have undefined here. If the `P` is inferred as `undefined`, it should be ok. Otherwise, it should be required to pass the props if there were some initially to avoid surprising results.

package.json
```diff
@@ -26,6 +27,7 @@
     "@babel/plugin-transform-modules-commonjs": "^7.2.0",
     "@babel/preset-env": "^7.3.4",
     "@babel/preset-react": "^7.0.0",
+    "@types/react": "^16.8.5",
```
FredyC: Where does it depend on React types actually? I don't see any reference to it.

otofu-square: It's necessary to use React hooks in typings test.
https://github.com/mpeyper/react-hooks-testing-library/pull/8/files#diff-e1558258b996f87650686aec9e520a38R1

index.d.ts
otofu-square: That's absolutely right.
Thanks for your suggestion!

index.d.ts
otofu-square: Looking at the implementation of the `rerender` function, it seems that `hookProps.current`(= `initialProps`) is assigned as a default value.
https://github.com/mpeyper/react-hooks-testing-library/blob/f7cd61035d8fb08cbfac376a5a8da1780708a64e/src/index.js#L26

As long as I see this, we can call `rerender` function with no argument or `initialProps`.

FredyC: I see, so it will use initialProps if none are passed. Makes sense I guess although it feels less explicit and might be confusing for tests.

mpeyper: Correct.  Subsequent rerenders will use the previously supplied props until new props are supplied. 

The idea is that the props don't change until the hook is rendered with new props, so hooks that use a conditional array (i.e. `useEffect`, `useLayoutEffect`, `useMemo` and `useCallback`) can be reliably tested.

mpeyper:
Ok, I feel like all comments have either been addressed or explained so I'm going to merge this.

@FredyC, @tatsushitoji Please open a PR adding yourself code review contributors by running `npm run contributors:add` and following the prompts.

PR #11, opened 2019-03-12
Add `waitForNextUpdate` promise for testing async updates

### LINKED ISSUE #10: Async hook testing
Hi, I am wondering how I can test hooks with async behavior.  I would like to create hooks that send a request to my api to rerender my component with some fresh data.

I tryied a couple of things and I couldn't find a way to test it. Is it supported by this library?

mpeyper:
**What**:

Provide ability to wait for hook to rerender to assist in testing asynchronous hooks (e.g. react-async's `useFetch`)

**Why**:

While asynchronous behaviour was not explicitly blocked by the library, testing it was often difficult when the internals of a hook used async patterns, but the underlying details (e.g. callback, `Promise`, etc.) were not exposed out of the hook's interface.

**How**:

The `renderHook` function now returns a `nextUpdate` function that returns a `Promise` that resolves when the hook is next rendered.  The signature and terminology is designed to work with the `async`/`await` pattern, e.g.

```js
test('something', async () => {
  const { result, nextUpdate } = renderHook(() => useFetch('/api'))

  expect(result.current.isLoading).toBe(true)

  await nextUpdate()

  expect(result.current.isLoading).toBe(false)
  expect(result.current.data).toEqual({/* ... */})
})
```

**Checklist**:

- [x] Documentation updated
- [x] Tests
- [ ] Ready to be merged
      
Before merging, I'd like some feedback on the interface, terminology and implementation of this feature.  Any feedback is welcome.

Resolves #10

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "contributors:add": "all-contributors add"
   },
   "dependencies": {
+    "@babel/runtime": "^7.3.4",
     "react-testing-library": "^6.0.0"
   },
   "devDependencies": {
@@ -34,6 +35,7 @@
     "@babel/core": "^7.3.4",
     "@babel/plugin-proposal-object-rest-spread": "^7.3.4",
     "@babel/plugin-transform-modules-commonjs": "^7.2.0",
+    "@babel/plugin-transform-runtime": "^7.3.4",
     "@babel/preset-env": "^7.3.4",
     "@babel/preset-react": "^7.0.0",
     "@types/react": "^16.8.5",
@@ -44,16 +46,13 @@
     "eslint": "^5.14.1",
     "eslint-config-prettier": "^4.0.0",
     "eslint-plugin-prettier": "^3.0.1",
-    "fetch-mock": "^7.3.1",
     "husky": "^1.3.1",
     "jest": "^24.1.0",
     "lint-staged": "^8.1.4",
-    "node-fetch": "^2.3.0",
     "prettier": "^1.16.4",
     "prettier-eslint": "^8.8.2",
     "prettier-eslint-cli": "^4.7.1",
     "react": "^16.8.3",
-    "react-async": "^5.1.0",
     "react-dom": "^16.8.3",
     "typescript": "^3.3.3333",
     "typings-tester": "^0.3.2"
--- a/src/index.js
+++ b/src/index.js
@@ -7,11 +7,8 @@ function TestHook({ callback, hookProps, children }) {
 }
 
 function renderHook(callback, { initialProps, ...options } = {}) {
-  const result = {
-    current: null
-  }
+  const result = { current: null }
   const hookProps = { current: initialProps }
-
   const resolvers = []
   const nextUpdate = () =>
     new Promise((resolve) => {
--- a/src/index.js
+++ b/src/index.js
@@ -10,7 +10,7 @@ function renderHook(callback, { initialProps, ...options } = {}) {
   const result = { current: null }
   const hookProps = { current: initialProps }
   const resolvers = []
-  const nextUpdate = () =>
+  const waitForNextUpdate = () =>
     new Promise((resolve) => {
       resolvers.push(resolve)
     })
@@ -28,7 +28,7 @@ function renderHook(callback, { initialProps, ...options } = {}) {
 
   return {
     result,
-    nextUpdate,
+    waitForNextUpdate,
     unmount,
     rerender: (newProps = hookProps.current) => {
       hookProps.current = newProps
--- a/src/index.js
+++ b/src/index.js
@@ -7,13 +7,22 @@ function TestHook({ callback, hookProps, children }) {
 }
 
 function renderHook(callback, { initialProps, ...options } = {}) {
-  const result = { current: null }
+  const result = {
+    current: null
+  }
   const hookProps = { current: initialProps }
+
+  const resolvers = []
+  const nextUpdate = () =>
+    new Promise((resolve) => {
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -44,16 +46,13 @@
...
-    "fetch-mock": "^7.3.1",
     "husky": "^1.3.1",
     "jest": "^24.1.0",
     "lint-staged": "^8.1.4",
-    "node-fetch": "^2.3.0",
     "prettier": "^1.16.4",
     "prettier-eslint": "^8.8.2",
     "prettier-eslint-cli": "^4.7.1",
     "react": "^16.8.3",
-    "react-async": "^5.1.0",
```
mpeyper: Whoops! these must have snuck into master when I was updating the README a few days ago. Nonetheless, they are not required.

typings/index.d.ts
```diff
@@ -9,6 +9,7 @@ export function renderHook<P, R>(
   readonly result: {
     current: R
   }
+  readonly nextUpdate: () => Promise<void>
```
mpeyper: not sure if `Promise<void>` is the most correct type here?

ab18556: Looks ok to me

ab18556:
I will not be able to test it until a couple of days, but it seems to be exactly what I needed ;)

src/index.js
```diff
@@ -9,11 +9,17 @@ function TestHook({ callback, hookProps, children }) {
...
+  const nextUpdate = () =>
+    new Promise((resolve) => {
+      resolvers.push(resolve)
+    })
 
   const toRender = () => (
     <TestHook callback={callback} hookProps={hookProps.current}>
       {(res) => {
         result.current = res
+        resolvers.splice(0, resolvers.length).forEach((resolve) => resolve())
```
ab18556: What if we pass res to resolve? Like this `resolvers.splice(0, resolvers.length).forEach((resolve) => resolve(res))`
Would it allow us to rename nextUpdate to nextValue and do assertions directly on it?
Like `expect(async nextValue).toEqual('Bob')`

typings/index.d.ts
ab18556: If we pass the value of the hook's result to nextUpdate/nextValue. I think that the signature should be `readonly nextValue: () => Promise<R>`

src/index.js
ab18556: What if we pass res to resolve? Like this `resolvers.splice(0, resolvers.length).forEach((resolve) => resolve(res))`
Would it allow us to rename nextUpdate to nextValue and do assertions directly on it?
Like `expect(async nextValue()).toEqual('Bob')`

mpeyper: I did start out with that, but decided that, as there is already a way of accessing the value, it might be confusing to use different variables before and after the `await` fold

```js
test('something', async () => {
  const { result, nextValue } = renderHook(() => useSomeHook())

  expect(result.current).toBe({/* ... */})

  expect(await nextValue()).toBe({/* ... */})
})
```

It also falls down when you want to assert a nexted value

```js
test('something', async () => {
  const { result, nextValue } = renderHook(() => useSomeHook())

  expect(result.current.someProp).toBe({/* ... */})

  // doesn't work
  // expect(await nextValue().someProp).toBe({/* ... */})

  const next = await nextValue()
  expect(next.someProp).toBe({/* ... */})
})
```

or multiple assertions

```js
test('something', async () => {
  const { result, nextValue } = renderHook(() => useSomeHook())

  expect(result.current).not.toContain({/* ... */})
  expect(result.current).toHaveLength(2)

  // fine
  // expect(await nextValue()).toContain({/* ... */})
  // hangs until the test fails
  // expect(await nextValue()).toHaveLength(3)

  const next = await nextValue()
  expect(next).toContain({/* ... */})
  expect(next).toHaveLength(3)
})
```

It also raises questions about whether it should fire if the `result.current` is set to the same value as it was already set to.  Consider a hook that has some internal state that updated asyncronously, triggering a side effect , but the hook returns a static result or no result at all itself (e.g. a `useRedux` hook that returns the static store, but a dispatch triggers some internal state to be updated).  If it didnt trigger on every render, this would once again be harder to test.

If we wanted to go this way, I think `result.nextValue()` _feels_ better to me for what it represents, but right now I still feel as the `nextUpdate` approach in this PR is more consistent, cleaner in more (common, in my opinion) use cases and requires fewer additional variables in complex tests.

ab18556: I do agree with you. Anyway, we can access the current right after we awaited so it doesn't really matter. 

About the method's name, I am wondering if it wouldn't be better to have something like the async methods in react-testing-library.

I am talking about the wait, waitForElement and waitForDomChange methods.

Maybe a name like waitForNextUpdate or waitForNextResult would provide a better description?

mpeyper: > waitForNextUpdate

I like that. `waitForNextResult` is fine but there are legitimate cases where a custom may not have a result at all (i.e. no return value), but `waitForNextUpdate` covers both nicely.

@ab18556 if you want to submit a PR [adding yourself as a contributor](https://github.com/mpeyper/react-hooks-testing-library/blob/master/CONTRIBUTING.md#add-yourself-as-a-contributor) (`review`, `ideas`, others?), I'll happily merge it.

ab18556: I don't fell i did that much, but since it was my first contribution to a public project, I am happy to do it.

mpeyper: You had more input on this PR than anyone else, and as a library author and maintainer, having anyone actually challenging and validating your work is invaluable and a fantastic contribution you can make (second only to updating the documentation 😉).

PR #16, opened 2019-03-19
Added global cleanup

### LINKED ISSUE #13: Add global cleanup
I love [react-testing-library](https://github.com/kentcdodds/react-testing-library) and would be interesting if [react-hooks-testing-library](https://github.com/mpeyper/react-hooks-testing-library) have the same clean up behavior.

So besides this

```js
import { cleanup } from 'react-hooks-testing-library';

afterEach(cleanup);
```

We could have one more way to do that and use the same approach as react-testing-library and use it in setup files

```js
import 'react-hooks-testing-library/cleanup';
// or
import 'react-hooks-testing-library/cleanup-after-each';
```

The implementation could be the same as react-testing-library

```js
// cleanup-after-each.js or just cleanup.js
afterEach(require('./dist').cleanup);

// to make sense to react-hooks-testing-library just change to /lib
afterEach(require('./lib').cleanup);
```

So the test suites would be more cleaner and you don't have to worry about cleanups. I would love to open a PR for this. 😁

marquesm91:
**What**:

Added global cleanup

**Why**:

To ease the cleanup tests using a global setup file

**How**:

Added in root `cleanup-after-each.js` to require `cleanup` function and call inside `afterEach`

**Checklist**:

- [ ] Documentation updated
- [ ] Tests
- [ ] Typescript definitions updated
- [X] Ready to be merged

I don't add anything to Documentation because `react-testing-library` don't. But if it is necessary I can complete [cleanup()](https://github.com/marquesm91/react-hooks-testing-library#cleanup) section with an example

Closes #13


mpeyper:
I should have read the description better as you explicitly mentioned documentation.

I think your suggestion of adding it to the [`cleanup()` section](https://github.com/marquesm91/react-hooks-testing-library#cleanup) is good.

marquesm91:
Added instructions in `cleanup` section. If you think it needs some more explatin let me know!

mpeyper:
Thanks @marquesm91.  I'll put together a release for this later (I really need to automate this).

jljorgenson18:
It looks like all of the cleanup functionality was removed. Is there any particular reason why?

mpeyper:
Hi @jljorgenson18,

Yes, we changed the renderer from `react-testing-library` to `react-test-renderer`.  One of the consequences of this change was that the test component we wrap the hook in is not rendered into a shared DOM anymore, so there is nothing to actually clean up between tests.

jljorgenson18:
Got it. Rendering though is only one facet to cleanup. If you have a side effect in useEffect (like a window event listener), would the cleanup function returned from useEffect get called?

mpeyper:
We never handled that explicitly (perhaps we should?). Usually you would `unmount` the hook yourself, triggering the cleanup, or handle your own specific cleanup yourself in a `beforeEach` or `afterEach` block.

Issue #19, opened 2019-03-21
Improve documentation


mpeyper:
The example in the README is very basic (by design) but there have been numerous queries recently about more advanced usage so I think it's time we expand the documentation to cover these.

Initially, this could be part of the README, but I'd prefer keeping the README fairly light and having a `docs` directory that can explain features, concepts and techniques in more details (a `.md` file per topic).

We could look at generating a [`gitbook`](https://www.gitbook.com/), [`docusaurus`](https://docusaurus.io/), [`docz`](https://www.docz.site/) or whatever the documentation site flavour of the month is from these docs at some point in the future (unless anyone want to set this up now).

Another thing that I am aware of ([from experience](https://ioof-holdings.github.io/redux-subspace/) 😞) is that docs written by the author are often not very useful to users (especially new users), as there is so much implied understanding that gets baked in.  That's why I'm asking for help for others to write these docs.  Include any gotchas or hiccups you hit along the way.

It doesn't have to be completely comprehensive from the get-go and we can improve and expand them as we go, so any and all contributions are welcome towards them.

Topics to cover:

- [x] when you should use this library
- [x] when you should *NOT* use this library
- [x] usage of `wrapper` (`renderHook` option)
  - [x] [context providers](#23)
- [x] usage of `initialProps` (`renderHook` option) and the `newProps` (`rerender` parameter)
  - [x] dynamic props
  - [x] memoization
- [x] usage of `waitForNextUpdate`
  - [x] async hooks
  - [x] suspence
  - [x] [warnings](https://github.com/mpeyper/react-hooks-testing-library/issues/14)
- [x] error handling
- [ ] recipies 
  - [ ] common hook patterns (if there are any?)
  - [ ] setup with different test runners
- [ ] examples
- [ ] troubleshooting
  - [x] ~[dom errors](https://github.com/mpeyper/react-hooks-testing-library/issues/17)~ (no more dom to cause errors (#40)
  - [ ] [result not changing](https://github.com/mpeyper/react-hooks-testing-library/issues/18)
- [x] FAQ
  - [x] [why do I need `.current`?](https://github.com/mpeyper/react-hooks-testing-library/issues/3)
- [x] API

There probably more that can also be included so feel free to go with want makes sense to you.  I'm happy to update the checklist if you want to make suggestions in the comments here.

[...]

PR #35, opened 2019-04-05
Add Suspense and ErrorBoundary to better support suspending from hooks

### LINKED ISSUE #27: Unable to test hooks that Suspend
- `react-hooks-testing-library` version: 0.4.0
- `react-testing-library` version: 6.0.3
- `react` version: 16.8.6
- `node` version: 11.10.0
- `npm` (or `yarn`) version: 1.15.2

### Relevant code or config:

```typescript
import { useResource } from 'rest-hooks';
renderHook(() => {
      article = useResource(CoolerArticleResource.singleRequest(), payload);
      return article;
    })
```

### What you did:

Running unit tests on 'rest-hook's library. Trying to update from react-testing-library - which works fine - to react-hooks-testing-library. 

(https://github.com/coinbase/rest-hooks/blob/master/src/react-integration/__tests__/integration.tsx)

### What happened:

Invariant Violation: Rendered more hooks than during the previous render.

### Reproduction:

https://github.com/coinbase/rest-hooks/blob/update-test/src/react-integration/__tests__/integration.tsx#L167
(run `yarn test` on that branch)

### Problem description:

Impossible to test hooks that use suspense. (This was fixed in React 16.8.2 https://github.com/facebook/react/pull/14821)

mpeyper:
**What**:

Added support for hooks that suspend during rendering.

**Why**:

We broke it when we started catching thrown values with the false assumption that they were always errors.

**How**:

1. Added a `Suspense` and `fallback` to allow the thrown promise to be handled
2. Moved the error handling into an `ErrorBoundary` to make it more "reacty"

**Checklist**:

- [ ] ~Documentation updated~
   - N/A - will make part of #19 
- [x] Tests
- [ ] ~Typescript definitions updated~
- [x] Ready to be merged
      
- [ ] ~Added myself to contributors table~
      

Fixes #27

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,12 +1,37 @@
-import React from 'react'
+import React, { Suspense } from 'react'
 import { render, cleanup, act } from 'react-testing-library'
 
 function TestHook({ callback, hookProps, children }) {
-  try {
-    children(callback(hookProps))
-  } catch (e) {
-    children(undefined, e)
+  children(callback(hookProps))
+  return null
+}
+
+class ErrorBoundary extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = { hasError: false }
   }
+
+  static getDerivedStateFromError() {
+    return { hasError: true }
+  }
+
+  componentDidCatch(error) {
+    this.props.onError(error)
+  }
+
+  componentDidUpdate(prevProps) {
+    if (this.props != prevProps && this.state.hasError) {
+      this.setState({ hasError: false })
+    }
+  }
+
+  render() {
+    return !this.state.hasError && this.props.children
+  }
+}
+
+function Fallback() {
   return null
 }
 
@@ -27,27 +52,34 @@ function resultContainer() {
     }
   }
 
+  const updateResult = (val, err) => {
+    value = val
+    error = err
+    resolvers.splice(0, resolvers.length).forEach((resolve) => resolve())
+  }
+
   return {
     result,
     addResolver: (resolver) => {
       resolvers.push(resolver)
     },
-    updateResult: (val, err) => {
-      value = val
-      error = err
-      resolvers.splice(0, resolvers.length).forEach((resolve) => resolve())
-    }
+    setValue: (val) => updateResult(val),
+    setError: (err) => updateResult(undefined, err)
   }
 }
 
 function renderHook(callback, { initialProps, ...options } = {}) {
-  const { result, updateResult, addResolver } = resultContainer()
+  const { result, setValue, setError, addResolver } = resultContainer()
   const hookProps = { current: initialProps }
 
   const toRender = () => (
-    <TestHook callback={callback} hookProps={hookProps.current}>
-      {updateResult}
-    </TestHook>
+    <ErrorBoundary onError={setError}>
+      <Suspense fallback={<Fallback />}>
+        <TestHook callback={callback} hookProps={hookProps.current}>
+          {setValue}
+        </TestHook>
+      </Suspense>
+    </ErrorBoundary>
   )
 
   const { unmount, rerender: rerenderComponent } = render(toRender(), options)
--- a/test/suspenseHook.test.js
+++ b/test/suspenseHook.test.js
@@ -1,0 +1,51 @@
+import { renderHook, cleanup } from 'src'
+
+describe('suspense hook tests', () => {
+  const cache = {}
+  const fetchName = (isSuccessful) => {
+    if (!cache.value) {
+      cache.value = new Promise((resolve, reject) => {
+        setTimeout(() => {
+          if (isSuccessful) {
+            resolve('Bob')
+          } else {
+            reject(new Error('Failed to fetch name'))
+          }
+        }, 50)
+      })
+        .then((value) => (cache.value = value))
+        .catch((e) => (cache.value = e))
+    }
+    return cache.value
+  }
+
+  const useFetchName = (isSuccessful = true) => {
+    const name = fetchName(isSuccessful)
+    if (typeof name.then === 'function' || name instanceof Error) {
+      throw name
+    }
+    return name
+  }
+
+  beforeEach(() => {
+    delete cache.value
+  })
+
+  afterEach(cleanup)
+
+  test('should allow rendering to be suspended', async () => {
+    const { result, waitForNextUpdate } = renderHook(() => useFetchName(true))
+
+    await waitForNextUpdate()
+
+    expect(result.current).toBe('Bob')
+  })
+
+  test('should set error if suspense promise rejects', async () => {
+    const { result, waitForNextUpdate } = renderHook(() => useFetchName(false))
+
+    await waitForNextUpdate()
+
+    expect(result.error).toEqual(new Error('Failed to fetch name'))
+  })
+})


src/index.js
```diff
@@ -1,12 +1,37 @@
...
 function TestHook({ callback, hookProps, children }) {
-  try {
-    children(callback(hookProps))
-  } catch (e) {
-    children(undefined, e)
+  children(callback(hookProps))
+  return null
+}
+
+class ErrorBoundary extends React.Component {
```
mpeyper: I wish there was a non-class way to handle error boundaries

test/suspenseHook.test.js
```diff
@@ -0,0 +1,51 @@
...
+
+  test('should allow rendering to be suspended', async () => {
+    const { result, waitForNextUpdate } = renderHook(() => useFetchName(true))
+
+    await waitForNextUpdate()
+
+    expect(result.current).toBe('Bob')
+  })
+
+  test('should set error if suspense promise rejects', async () => {
```
mpeyper: I'm not sure this test actually adds much value.  in #27 I discovered that a rejected `Promise` does not trigger the error boundary, and instead it's up to the suspending code to throw the error on the subsequent render (see line 24 for example), which means this is actually no different than if the hook just threw an error for some other reason (like we already do in the `errorHook` test suite.

mpeyper:
@ntucker Please submit a PR [adding yourself as a contributor](https://github.com/mpeyper/react-hooks-testing-library/blob/master/CONTRIBUTING.md#add-yourself-as-a-contributor).  You definitely get the `bug` contribution for #27, but I feel like you also need something for the help you gave me in getting my test cases set up properly (`test` or `examples`?).  If you want to review this PR you can have a `review` contribution too 😉.

src/index.js
```diff
@@ -1,12 +1,37 @@
-import React from 'react'
+import React, { Suspense } from 'react'
 import { render, cleanup, act } from 'react-testing-library'
 
 function TestHook({ callback, hookProps, children }) {
-  try {
```
ntucker: If you kept this block an just re-threw you could also grab any promises to store in a third result value of 'promise' or the like. That would make testing suspense much easier because then we could await that. At the very least inspecting the promise thrown would be useful in tests.

ntucker: (If you rethrow the exact same error it will preserve the original stack trace.)

test/suspenseHook.test.js
ntucker: Anyone who implements a Suspense library will be doing such so I think this provides value

mpeyper: Yeah, that was my thought to and why I kept it when clean up the commits.

src/index.js
mpeyper: The intention of removing this and using a `Suspense` and `ErrorBoundary` wrapper was to protect this library from having to deal with the internals of how error handling and suspended rendering actually works.  The fact that a thrown promise is what triggers react to suspend is an implementation for React, rather than something we should be relying on.

ntucker: you're still setting error property though, right? Seems like you are somewhat depending on react in that manner. regardless - this *is* a react hooks testing library. very specific to how react works.

mpeyper: Yes, `error` is being set, but it's coming from the exposed `componentDidCatch` API, so we don't care _how_ React gets it to that point, just deal with it when it gets there.  When we look at the `Suspense` equivalent, React does not expose this to us in any way, so if we want to handle it, we essentially need to recreate react's internal logic in order to treat it the same way.  If React changes the specifics on how it deals with promises, we would also need to change.

mpeyper: But I am interested to see an example of when you would use a captured error that is not covered by `waitForNextUpdate` and `result.current`.  IF the use case is strong enough, then perhaps the additional maintenance burden is worth it.  I just don't see it yet.

ntucker: Currently I'm wrapping everything in suspense so I can track when it gets suspended. Suspense triggering is a very common part of my tests. That can still be done with the fallback prop. I won't need to look at the exact promise if `waitForNextUpdate()` works, but still need:

- the promise to be caught so it doesn't throw above
- a mechanism to see when it is called

mpeyper: > if waitForNextUpdate() works

It should.  I'd rather fix holes where it doesn't than custom suspense handling.

> - the promise to be caught so it doesn't throw above

This solution would not end up throwing out of the `renderHook` function, if that's what you mean?

> - a mechanism to see when it is called

Can you unpack this a bit more?  Are you concerned about when an update was triggered by something other than the promise resolving, or something else?

ntucker: One path is that no promise is thrown from the hook - I want to test that it actually threw a promise, and THEN did something else

mpeyper: I guess I'm questioning the value of having that much pwoer to inspect the internals of what's happening in your hook (i.e. white box test) vs. just asserting the output from the given inputs (i.e. black box test).

In my experience, white box test are very fragile to changes in the system and need constant updating as the internals shift around (e.g. you might change the condition for _when_ the promise is thrown, so now would need to move things around in your tests to match the change).  The black box tests are more likely to continue to work so, long as the contract of the inputs and outputs hasn't changes (otherwise the test _must be_ changed anyway in both testing styles).

In cases where you need more control over when and how the promise thrown, then you may be better off extracting it into some form of factory pattern and mocking that behaviour in the hook test, and testing those variations of the factory seperately to the hook itself.

As an escape hatch, if you _really_ need access to the promise itself, I think you could just catch it in the hool callback itself

```js
let promise
  promise = null // assuming you want to reset it between renders
  try {
    useSomething()
  } catch (e) {
    if (e.then) {
      promise = e
    }
  }
})

const promiseValue = await promise

expect(promiseValue).toBe('whatever')
```

I'm still unsure how that is more useful than

```js
const { result, waitForUpdate } = renderHook(() => useSomething())

await waitForUpdate

expect(result.current).toBe('whatever you expect after the prmise resolved')
```

As in both cases
* if the promise resolves, and the expectation is met, the the test passes
* if the promise resolves, and the expectation is not met, the test fails
   * in the first example all you would know the promise value changed
   * in the second example you would only know something went wrong to affect the output
* if the promise rejects, the test fails with the thrown error
* if the promise is not thrown, the test will fail
   * in the first example it would probably be about trying to `await` `null`
   * in the second example the test would timeout waiting for an update that never comes

mpeyper: I'm also leaning away from explicitly handling supsense as it would get very difficult to manage if you had a hook throw a promise, and then once resolved throw another promise, etc. Suspense would keep that component suspending untill it eventually rendered something, whereas keeping track of each one and what they were up to could get complicated.

mpeyper:
@ntucker I'm going to merge this as-is for now as I'm aware that no suspense compatible hooks can be tested int he current version and what we are discussing would probably be additive to this change anyway.  Please feel free to raise a new issue referencing this PR and #27 if you want to discus it further.

Also, I'm still waiting for [that PR to add yourself as a contributor](https://github.com/mpeyper/react-hooks-testing-library/pull/35#issuecomment-480246413).  If you don't want to do it, that's fine, or if you're happy, I can make the addition for you (but you won't get the github cred if I do it).

src/index.js
ntucker: I mean, I would argue that suspense is very much an explicit interface and not an implementation detail - as anyone using the hook that throws promises will need to define suspense boundaries themselves. It's not about *how* but *if*.

Yes it could throw many times - that should be up to the test writer to understand exactly what that behavior should be - or if it even cares at all. I agree in many cases you won't care, and you can simply waitForUpdate. But there are certainly cases where you will

ntucker:
Ya that seems reasonable. This is definitely an improvement that should be released for people to use.

PR #47, opened 2019-04-22
Fix example

josepot:
**What**:

I've fixed the code of the main example.

**Why**:

Because as I explain in [this post](https://dev.to/josepot/react-s-bogus-setstate-anti-pattern-287e), the custom-hook of the main example wouldn't work correctly otherwise.

**How**:

I just used the function overload of `setState`.

**Checklist**:

Yep

- [X] Documentation updated
- [X] Ready to be merged


mpeyper:
Nice observation.

Issue #76, opened 2019-05-17
Global `unmountAll` functionality


DaveStein:
I have an effect that uses `setInterval`. In order for my tests to work right, I need to `unmount` so the interval gets knocked away.

I have code like this:

```js
let globalUnmount;

// I use this is all of my tests to run render the hook
async function runEffects(wait = true) {
  const { result, unmount } = renderHook(() =>
    useMyHookWithIntervals()
  );

  globalUnmount = unmount;
  return {
    result
  };
}
afterEach(() => {
  if (globalUnmount) {
    globalUnmount();
    globalUnmount = null;
  }
});
```

Is this the expected paradigm? Is this an example that would be good for docs? Is this horrible?

jljorgenson18:
We have the same problem. When this library moved to react-test-renderer, the cleanup function was removed. This is fine when you are just testing rendering, but a big part of hooks is side effects. There doesn't seem to be a great way to have a global unmount and we are currently stuck on v0.4.1.

userbq201:
+ 1. Unmount doesnt work

DaveStein:
@userbq201 if you are trying to report a bug, you should probably open a different ticket. This ticket is not about a bug, just about how to best use what features are available.

mpeyper:
Hey, sorry I've been a bit quiet.  I've been moving interstate so my spare time has been a bit light on, but I'm here now.

To be honest, an `unmountAll` is not something I've considered.  The snippet you've provided would be a sensible way to do it with the current API.

If anyone wants to implement something like that and submit a PR, I think it would be a useful addition to the library.  It would need to track each rendered hooks and unmount all of them, unlike this snippet which only unmounts the last rendered hook, but that should not be too difficult.

mpeyper:
Just wondering if this should be called `unmountAll` or if we bring back `cleanup` to give us room to add more cleanup steps in the future?

DaveStein:
I like how jest has the different levels of `clear` vs `reset` mocks and whatnot. So being specific now, and allowing for other specific cleanup steps in the future would seem good. Then we can always have a `cleanup` that combines all as needed.

apostolidhs:
A quick workaround for me was to create a small wrapper and use only this for the tests

```javascript
//  './renderHook.js';
import {renderHook} from '@testing-library/react-hooks';

let unmounts = [];

export const cleanup = () => {
  unmounts.forEach(unmount => unmount());
  unmounts = [];
};

[...]

PR #106, opened 2019-06-24
Pr/cleanup renderhooks

apostolidhs:
**What**:

Added global cleanup

**Why**:

We should unmount all the rendered hooks after each test so we can avoid memory leaks, flaky tests and fully test the hooks on the entire lifecycle.

**How**:

We keep the functions that unmount of the rendered hooks in a list and export a `cleanup` function that iterates the list and unmounts the hooks.

**Checklist**:

- [x] Documentation updated
- [x] Tests
- [ ] Ready to be merged
      
I don't know if we should add a global option on the library that enables the cleanup feature. If a client runs the tests without calling the cleanup function, then we will collect a list of the `unmount` functions. This will may cause performance issues in a huge amount of tests.

---

Related issue https://github.com/testing-library/react-hooks-testing-library/issues/76


src/index.js
```diff
@@ -51,6 +53,11 @@ function resultContainer() {
   }
 }
 
+function cleanup() {
+  unmounts.forEach((unmount) => unmount())
+  unmounts.length = 0
```
mpeyper: is there significant difference between this and `unmounts = []`?  That feels a bit clearer what it's doing to me, but I'm not sure if there are some other hidden benefits to this approach?

src/index.js
```diff
@@ -73,6 +80,14 @@ function renderHook(callback, { initialProps, wrapper } = {}) {
   })
   const { unmount, update } = testRenderer
 
+  function unmountHook() {
```
mpeyper: what happens if someone calls this then then `cleanup`?  does the second `unmount` fail?  should `unmountHook` also remove the callback from `unmounts` (making [this line](https://github.com/testing-library/react-hooks-testing-library/pull/106/files#r296980396) no longer necessary at all)

cleanup-after-each.js
```diff
@@ -0,0 +1 @@
+afterEach(require('./src').cleanup)
```
mpeyper: I think this should be requireing `./lib` so that it picks up the compiled version after publishing.

src/index.js
mpeyper: I see there is a test around this case.  I personally feel that having `unmountHook` remove the callback from the `unmounts` array is bit clearer to what is happening rather than relying on the behaviour of `unmount`, but I don't care enough to block this for it.

mpeyper:
Hi @apostolidhs,

Thanks for taking the time to put this together.  There was still some discussion in #76 about whether this should be called `cleanup` or `unmountAll`.  I personally prefer `cleanup`, but I'll give others the opportunity to change my mind before we commit to this.  @DaveStein feel free to weigh in on this one.

src/index.js
apostolidhs: I was thinking to use either mutable or immutable approach. Personally, i prefer the immutable way as i suggested in the issue.

```
let unmounts = [];
unmounts = [...unmounts, unmount];
unmounts = [];
```

but i saw the way that you are using the `resolvers`

```
const resolvers = []; 
resolvers.push(resolver);
``` 
and i tried to follow the same pattern for consistency (mutable). I don't have a strong opinion on this, it's up to you.

src/index.js
apostolidhs: I agree with you, i will fix it.

mpeyper:
I just read the description properly and noticed your question about a global option.

Exploring that a bit, would it be enabled by default and they can disable globally if they have performance issues or is it disabled by default and they have to opt in to use it? We could enable it by default if they use the `cleanup-after-each` utility which would cover most use cases (in my opinion).

I also think it would take an awful lot of tests before someone actually saw an issue from not calling `cleanup`, so perhaps we don't worry until someone files an issue?

mpeyper:
Just another thought, could we auto clear on a `renderHook` call? My test generally only have 1 to kick things off, then I use `rerender` and `act` calls for follow up renders. 

Would this result in less cases where anyone actually needs to call `cleanup` or use the `cleanup-after-each` utility?

Would this result in more strange failures because something throwing on unmount would fail the next test?

DaveStein:
My comment about `unmountAll` vs `cleanup` would just be a reference to how Jest has `reset` vs `clear` and the like. We might want `cleanup` to do all the things, and have `unmountAll` to do one form of cleanup. Otherwise, someone might mean to just `unmountAll` in tests today, then have something added to `cleanup`, and then have their tests break because it's doing something extra.

apostolidhs:
@mpeyper Can you take a look again? I committed the changes as a fixup. 

* Renames the cleanup -> unmountAll
* Remove the unmounted hook from the unmounts

mpeyper:
Sorry, I haven't had time to looks at this again yet (end of financial year is Oz and I work in financial services). I'm planning to look in the next few days.

mpeyper:
@apostolidhs Those changes look good.  I found them [here]( https://github.com/apostolidhs/react-hooks-testing-library/commit/13ba272f15fd86ac7a6a7a14be9ac51c7aa0bc85), but they do not appear to be in this branch (I did not the branch they are on is missing the `pr/` prefix that this branch has).  Can you add that commit to this PR?

The only bit I'm not sure on is where you've put the documentation (in the README). It would be great to get it into the [docs site](https://react-hooks-testing-library.com/setup).  Perhaps more or less what you have in the [setup page](https://react-hooks-testing-library.com/setup) and then some details on the `unmountAll` export in the [API reference](https://react-hooks-testing-library.com/reference/api).

I can tidy all that up after we merge this if you prefer too.  Just let me know.

mpeyper:
I'm not sure where this is at right now, but react-testing-library has a PR open around auto cleanup, which might be something we want to explore instead of/in addition to this.

https://github.com/testing-library/react-testing-library/pull/430

mpeyper:
I'm going to close this as it appears to have gone stale (no updates in months and 4/6 changed files not have conflict.

I'm still very much in favour of this change, so if anyone wants to pick it up from here then I'm happy to continue.  Alternatively, if you want to start fresh, that's cool too.

PR #111, opened 2019-06-28
Make each interface public

VinceMalone:
**What**:

Make each interface public for use in explicit types, rather than relying on the implicit typing from `renderHook`.

**Why**:

Implicit types often work just fine, but I've run into a case where I want to reference the typing (related to `renderHook`) in something like a shared utility.

For example:

```ts
import { RenderHookResult, renderHook } from '@testing-library/react-hooks';

const results = new Set<RenderHookResult<unknown, void>>();
```

**How**:

By exporting each `interface`

**Checklist**:

- [x] Ready to be merged


mpeyper:
Looks good. I don't have time to do a publish a release at the moment (end of financial year in Oz and I work in financial services) so I'll merge this in a few days.

Issue #110, opened 2019-06-26
Duplicated identifier Errors with TypeScript


fxOne:
- `react-hooks-testing-library` version: 1.0.4
- `react-test-renderer` version: 16.8.6
- `react` version: 16.8.17
- `@types/react` version: 16.8.17
- `typescript` version: 3.3.3333
- `node` version: 11.9.0
- `npm`  version: 6.7.0

### Relevant code or config:

\-

### What you did:

I installed the latest Version of `react-hooks-testing-library` and got after a `typecheck` the following error.

### What happened:

```
node_modules/@testing-library/react-hooks/node_modules/@types/react/index.d.ts:2817:14 - error TS2300: Duplicate identifier 'LibraryManagedAttributes'.

2817         type LibraryManagedAttributes<C, P> = C extends React.MemoExoticComponent<infer T> | React.LazyExoticComponent<infer T>
                  ~~~~~~~~~~~~~~~~~~~~~~~~

  node_modules/@types/react/index.d.ts:2816:14
    2816         type LibraryManagedAttributes<C, P> = C extends React.MemoExoticComponent<infer T> | React.LazyExoticComponent<infer T>
                      ~~~~~~~~~~~~~~~~~~~~~~~~
    'LibraryManagedAttributes' was also declared here.

node_modules/@testing-library/react-hooks/node_modules/@types/react/index.d.ts:2881:13 - error TS2717: Subsequent property declarations must have the same type.  Property 'iframe' must be of type 'DetailedHTMLProps<IframeHTMLAttributes<HTMLIFrameElement>, HTMLIFrameElement>', but here has type 'DetailedHTMLProps<IframeHTMLAttributes<HTMLIFrameElement>, HTMLIFrameElement>'.

2881             iframe: React.DetailedHTMLProps<React.IframeHTMLAttributes<HTMLIFrameElement>, HTMLIFrameElement>;
                 ~~~~~~

node_modules/@types/react/index.d.ts:2816:14 - error TS2300: Duplicate identifier 'LibraryManagedAttributes'.

2816         type LibraryManagedAttributes<C, P> = C extends React.MemoExoticComponent<infer T> | React.LazyExoticComponent<infer T>
                  ~~~~~~~~~~~~~~~~~~~~~~~~

  node_modules/@testing-library/react-hooks/node_modules/@types/react/index.d.ts:2817:14
[...]

PR #113, opened 2019-07-10
Move type dependencies to peer dependencies.

ericmackrodt:
**What**:

Moves the `@types` from `dependencies` to `peerDependencies` in the package.json.

**Why**:

Because the dependencies are always installed, if the project this library is being used already has references to those types but in a different version, especially if you're using Lerna, you get multiple `Duplicate identifier` errors in typescript.

**How**:

The dependencies were moved to `peerDependencies` and also added to `devDependencies` for the library development.

**Checklist**:

- [x] Documentation updated
- [x] Tests
- [ ] Ready to be merged
      
### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -120,6 +120,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "ericmackrodt",
+      "name": "Eric Mackrodt",
+      "avatar_url": "https://avatars3.githubusercontent.com/u/450616?s=180&v=4",
+      "profile": "http://ericmackrodt.com",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "commitConvention": "none"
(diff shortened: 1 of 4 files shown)


mpeyper:
This has been discussed a bit in #110 already, but I'm not convinced the type dependencies are actually wrong here.  Because of semver versioning, the types should resolve to a single version, so long as they are compatible semver ranges.

In the case of lerna, `npm link` (which lerna uses under the hood) is [well documented](https://github.com/Microsoft/typescript/issues/6496) as causing issues when resolving type definitions.

It's an annoying situation for me because either I make the type definitions a `dependency` and have it install the right one for all the cases when someone doesn't care about they type definition (i.e. non-typescript user or typescript user without that dependency themselves), or I have them as `peerDependencies` and make everyone who doesn't care about them get a `peerDependency` warning on installation (less important for a typescript user without that dependency, as they should be installing it anyone, but a pain for a non-typescript user).

mpeyper:
Closing this.  See #110 for details.

Issue #135, opened 2019-08-03
Get how many times the component was rendered


Belco90:
### Describe the feature you'd like:
I would like to get inside `result` a property with number of times the component was rendered. It could be called `renderCount`.

The only drawback I can think about is performance issues, but I don't think a counter is a problem.

### Suggested implementation:

One option could be wrapping the `setValue` passed to `TestHook` so it increase an external count each time than `setValue` is called.

Another option could be to pass an utils object to `TestHook` as a `children` with properties `setValue` and `increaseCount` so `TestHook` itself is in charge of increasing the count before calling `children.setValue`.

I think it's important increasing the count even if it fails, as in the end the component was triggered to be rendered.

Not really sure about where to init the count. Maybe a ref inside `TestHook`?

### Describe alternatives you've considered:
I'm trying to test a hook that returns a function to force a component to be re-rendered. I've tried to pass the count from outside but 1) I can't pass inner components, only outer (the wrapper) or 2) pass component prop but this is not possible.

I could build my own testing component with react-testing-library that calls the function returned by the hook when a button is clicked or something like that but I only have this library available in the project for now and I don't want to install RTL only for this test.

I can't think other way to test the hook I described.

### Teachability, Documentation, Adoption, Migration Strategy:
I think is useful to get how many times the component was rendered for other purposes too: checking performance issues or memoized behaviors.

Would be easy to include in the docs, just adding a new element to renderHook Result section explaining what do you get for `renderCount` property.

Belco90:
I just figured out how this can be implemented and created proper PR https://github.com/testing-library/react-hooks-testing-library/pull/136

mpeyper:
Closing for now.  [See comment in the PR](https://github.com/testing-library/react-hooks-testing-library/pull/136#issuecomment-519894745) for details.

TheNotary:
I'm using useContext for the first time.  Things have gotten pretty complex with it due to my noviceship and I needed a way to add a simple count to my `<LoginPage />` which is really just a testbed page at this point specific to Logins and the userContext.  Because of interactions between my javascript Proxy object and setState, I was getting ridiculous amounts of re-rendering taking place and odd bugs related to creating proxies of proxies.  

I thought it was going to be an excruciating challenge doing this in a react environment, but it wound up being really easy.  I just render a `let` variable that's instantiated outside of my component, and incremented within the component.  Now I'll know if I re-introduce the bug because the number will make really large jumps as I interact with the page.  

```
[...]

PR #136, opened 2019-08-03
Add render count

Belco90:
**What**:
Implementing feature suggested to get how many times the component has been rendered https://github.com/testing-library/react-hooks-testing-library/issues/135

**Why**:
At the moment there is no way of knowing how many times a component has been rendered due to a hook update, and it's an interesting value for checking performance, memoized behaviors or forced updates.

**How**:
A count init to 0 within `resultContainer` that is increased every time `updateResult` is called, no matter if there is an error executing the hook as another render has been triggered so the counter must be increased to get corresponding count.

**Checklist**:

- [x] Documentation updated
- [x] Tests
- [x] Ready to be merged
      
### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -120,6 +120,17 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "Belco90",
+      "name": "Mario Beltrán",
+      "avatar_url": "https://avatars1.githubusercontent.com/u/2677072?v=4",
+      "profile": "https://mario.dev",
+      "contributions": [
+        "code",
+        "doc",
+        "test"
+      ]
     }
   ],
   "commitConvention": "none"
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@
 [![downloads](https://img.shields.io/npm/dm/@testing-library/react-hooks.svg?style=flat-square)](http://www.npmtrends.com/@testing-library/react-hooks)
 [![MIT License](https://img.shields.io/npm/l/@testing-library/react-hooks.svg?style=flat-square)](https://github.com/testing-library/react-hooks-testing-library/blob/master/LICENSE.md)
 
-[![All Contributors](https://img.shields.io/badge/all_contributors-11-orange.svg?style=flat-square)](#contributors)
+[![All Contributors](https://img.shields.io/badge/all_contributors-12-orange.svg?style=flat-square)](#contributors-)
 [![PRs Welcome](https://img.shields.io/badge/PRs-welcome-brightgreen.svg?style=flat-square)](http://makeapullrequest.com)
 [![Code of Conduct](https://img.shields.io/badge/code%20of-conduct-ff69b4.svg?style=flat-square)](https://github.com/testing-library/react-hooks-testing-library/blob/master/CODE_OF_CONDUCT.md)
 [![Netlify Status](https://api.netlify.com/api/v1/badges/9a8f27a5-df38-4910-a248-4908b1ba29a7/deploy-status)](https://app.netlify.com/sites/react-hooks-testing-library/deploys)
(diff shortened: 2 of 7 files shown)


mpeyper:
Hi @Belco90,

Thanks for taking the time to put this together and the associated issue (#135).

I'm in two minds about this feature.  My one half is thinking that this is a simple enough feature and to accept this, but my other half is worried that it's not quite in line with the spirit of the `@testing-library` packages.  Let me explain...

I think the benefits of it is a bit more limited than you "Why" section describes with respect to testing for memoization.  A hook will rerender the component on any state changes, at which point it can use memoised values to optimise its own performance, but the render still occurs.  Unless you are testing the identity of some returned values to ensure they haven't changed, the test itself would be oblivious to whether memoization was used.

So this leaves us with testing performance and forced updates.

I'll tackle forced updates first as it's a bit easier to explain.  Basically, it is very rare that you actually want to force an update in `react`.  Generally speaking, the changing state of a component should be what triggers any updates and needing to force an update is usually a sign of something else not behaving correctly in the app.  That said, we already provide the `rerender` function that will rerender the component and allow the hook's result to be recalculated if required.

The performance point is where the discussion gets interesting for me.  On the surface, I'd say that testing for performance is a good thing.  I'm just not sure using render counts is the best way to do this.  I feel that a change like this will encourage users to focus on ensuring the component renders once, _and only once_, on any given action and spend a lot of time on micro-optimisations when, in reality, just letting `react` rerender the component is often good enough for the majority of use cases.

Essentially, it feels a little bit like we're getting bogged down in the details of _how_ the result was calculated, rather than focussing on the correctness of the result itself.  It would be easy to validly refactor the hook to render more or fewer times but still produce the same result, causing the test to now fail due to an implementation detail.

So I guess my objection is coming from a desire to help the users **not** think about any of this at the expense of a small number of valid use cases.  In those cases, it is possible to provide render counts without changing this library, e.g:

```js
  function useUpdate() {
    const [, setState] = useState(0)
    return () => setState((count) => count + 1)
  }

  test('should increase render count every time hook update function is called', () => {
    let renderCount = 0
      renderCount++
      return useUpdate()
    })
    const update = result.current

    expect(renderCount).toBe(1)

    act(() => update())
    expect(renderCount).toBe(2)

    act(() => update())
    expect(renderCount).toBe(3)
  })
```

If you have lots of tests you want need to use this for a wrapper around `renderHook` can make it a bit more generic.  Something like:

```js
function renderHookWithCount(hookCallback) {
  const container = renderHook((...props) => {
    if (container) {
      container.result.renderCount++;
    }
    return hookCallback(...props);
  });
  container.result.renderCount = 1;
  return container;
}

  test('should increase render count every time hook update function is called', () => {
    const { result } = renderHookWithCount(() => useUpdate())
    const update = result.current

    expect(result.renderCount).toBe(1)

    act(() => update())
    expect(result.renderCount).toBe(2)

    act(() => update())
    expect(result.renderCount).toBe(3)
  })
```

All this being said, I'll think about this a bit more and listen to any feedback you or anyone else wants to share here about my opinions or other possible use cases this feature might enable and make a decision in the next day or two.

One thing to think about in the meantime is what should the `renderCount` be for a component that suspends in it's first render and completes in its seconds render?  Again I'm torn on this.  Part of me says it's obviously 2 renders , but the other part says it's 1 completed render.  Right now I think your implementation will have `renderCount` as 1 in this scenario [because we don't update the result when a promise is thrown](https://github.com/testing-library/react-hooks-testing-library/blob/master/src/index.js#L8-L10).

Belco90:
Aaah of course! Didn't think about increasing the count inside `renderHook` callback.

About the "why" of this feature implementation: you are right, it doesn't seem like following patterns from `@testing-library` and at the end the _only_ interesting use case is performance.

Didn't think about the render count for suspend, I guess that makes things more complicated as my original proposal.

This is quite interesting:
> Essentially, it feels a little bit like we're getting bogged down in the details of how the result was calculated, rather than focussing on the correctness of the result itself. It would be easy to validly refactor the hook to render more or fewer times but still produce the same result, causing the test to now fail due to an implementation detail.

So probably for performance this feature only makes sense as a profiling/measurement rather than a value for testing itself. But it looks like is the first time this has been mentioned (and I'm little bit surprised!), so probably all this is not necessary at all.

Your workaround solves my initial problem, but I'm happy to help implementing or discussing whatever is needed if you find this feature interesting for performance measurement.

mpeyper:
I'm going to close this without merging for now.  I'm still of the opinion that this is not needed and a variation of the wrapper above will work for most use cases for now.  If there is a big demand in the future, we can revisit this then.

sultan99:
Another way to count renders:

```js
const renderHookWithCount = hook => renderHook(() => {
  const countRef = useRef(0)
  countRef.current ++
  return {renderCount: countRef.current, ...hook()}
})

const {result} = renderHookWithCount(() => ({
 someValue: someCustomHook()
})

expect(result.current.someValue).toBe(`some-value`)
expect(result.current.renderCount).toBe(1)
```

PR #199, opened 2019-10-14
Global cleanup

### LINKED ISSUE #76: Global `unmountAll` functionality

mpeyper:
**What**:

Added cleanup functionality to unmount any rendered test components.

Resolved #76  

**Why**:

Currently, the underlying rendered test component is not necessarily unmounted at the end of a test, resulting in effects from `useEffect` and `useLayoutEffect` hanging around between tests.

**How**:

The library now exports a global `cleanup` function that will unmount any rendered hooks.  As per [testing-library/react-testing-library#430](https://github.com/testing-library/react-testing-library/pull/430) we also auto regester an `afterEach` handler in supported test frameworks (with an environment variable opt out).

**Checklist**:

- [x] Documentation updated
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -9,6 +9,7 @@ route: '/reference/api'
 
 - [`renderHook`](/reference/api#renderhook)
 - [`act`](/reference/api#act)
+- [`cleanup`](/reference/api#cleanup)
 
 ---
 
@@ -102,3 +103,23 @@ A function to unmount the test component. This is commonly used to trigger clean
 
 This is the same [`act` function](https://reactjs.org/docs/test-utils.html#act) that is exported by
 `react-test-renderer`.
+
+---
+
+## `cleanup`
+
+Unmounts any rendered hooks rendered with `renderHook`, ensuring all effects have been flushed.
+
+> Please note that this is done automatically if the testing framework you're using supports the
+> `afterEach` global (like mocha, Jest, and Jasmine). If not, you will need to do manual cleanups
+> after each test.
+>
+> Setting the `RHTL_SKIP_AUTO_CLEANUP` environment variable to `true` before the
+> `@testing-library/react-hooks` is imported will disable this feature.
+
+```js
+async function cleanup: void
+```
+
+The `cleanup` function should be called after each test to ensure that previously rendered hooks
+will not have any unintended side-effects on the following tests.
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -108,6 +108,10 @@ This is the same [`act` function](https://reactjs.org/docs/test-utils.html#act) 
 
 ## `cleanup`
 
+```js
+function cleanup: Promise<void>
+```
+
 Unmounts any rendered hooks rendered with `renderHook`, ensuring all effects have been flushed.
 
 > Please note that this is done automatically if the testing framework you're using supports the
@@ -116,10 +120,6 @@ Unmounts any rendered hooks rendered with `renderHook`, ensuring all effects hav
 >
 > Setting the `RHTL_SKIP_AUTO_CLEANUP` environment variable to `true` before the
 > `@testing-library/react-hooks` is imported will disable this feature.
-
-```js
-async function cleanup: void
-```
 
 The `cleanup` function should be called after each test to ensure that previously rendered hooks
 will not have any unintended side-effects on the following tests.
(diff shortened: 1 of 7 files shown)


src/cleanup.js
```diff
@@ -0,0 +1,26 @@
...
+function addCleanup(callback) {
+  cleanupCallbacks.push(callback)
+}
+
+function removeCleanup(callback) {
+  cleanupCallbacks = cleanupCallbacks.filter((cb) => cb !== callback)
+}
+
+// Automatically registers cleanup in supported testing frameworks
+if (typeof afterEach === 'function' && !process.env.RHTL_SKIP_AUTO_CLEANUP) {
```
hartzis: 📓 `process.env.RHTL_SKIP_AUTO_CLEANUP` only needs to be "truthy" and not `true` to work.

Do we want this to explicitly be the string `"true"`?

docs/api-reference.md
```diff
@@ -102,3 +103,23 @@ A function to unmount the test component. This is commonly used to trigger clean
...
+
+```js
+function cleanup: Promise<void>
+```
+
+Unmounts any rendered hooks rendered with `renderHook`, ensuring all effects have been flushed.
+
+> Please note that this is done automatically if the testing framework you're using supports the
+> `afterEach` global (like mocha, Jest, and Jasmine). If not, you will need to do manual cleanups
+> after each test.
```
hartzis: ❓ Do you happen to have any reference or example of a testing library auto adding a global `afterEach`?

I think I like this, i'm just curious about community best practices and such.

test/autoCleanup.noAfterEach.test.js
```diff
@@ -0,0 +1,28 @@
...
+        return () => {
+          cleanupCalled = true
+        }
+      })
+    }
+    renderHook(() => hookWithCleanup())
+  })
+
+  test('second', () => {
+    expect(cleanupCalled).toBe(false)
```
hartzis: ❓ Is this test supposed to check `true`, `expect(cleanupCalled).toBe(true)`?

It appears to be a copy of the above test, `autoCleanup.disabled.test.js`, but without the `process.env` setup.

You may need to clear the `require` cache and reset the `process.env` before running?

docs/api-reference.md
mpeyper: The registration of this is lifted almost verbatim from [`@testing-library/react-testing-library#430`](https://github.com/testing-library/react-testing-library/pull/430)

I don't think this is a commonly used pattern (yet).

src/cleanup.js
mpeyper: Again, this is implemented as per [`@testing-library/react-testing-library`](https://github.com/testing-library/react-testing-library/blob/master/src/index.js#L9)

test/autoCleanup.noAfterEach.test.js
mpeyper: No.  In the `beforeAll` am overriding `afterEach` to not be a function to test the case where the testing framework does not support it.  I was very worried about this, but I tested thoroughly and `jest` brings it back for the next test.

Happy to hear ideas for better ways to test this case.

hartzis: welp `jest` is pretty damn cool.

mpeyper:
Yeah, will have to be a major, even though I doubt anyone will actually be affected by the break.

mpeyper:
Ok, I'm going to merge this now and get a major release out.

@hartzis, if you would like to, you are welcome to submit a PR [adding yourself as a contributor](https://github.com/testing-library/react-hooks-testing-library/blob/master/CONTRIBUTING.md#add-yourself-as-a-contributor) for your efforts reviewing this PR.

PR #200, opened 2019-10-14
wait and waitForValueToChange async utils

### LINKED ISSUE #173: Warning about using different versions of act() when wait()ing on an async effect
- `@testing-library/react-hooks` version: 2.0.1
- `react-test-renderer` version: 16.9.0
- `react` version: 16.9.0
- `node` version: 12.9.0
- `npm` (or `yarn`) version: 1.17.3 (yarn)

### Relevant code or config:

```js
import { useState, useCallback, useEffect } from 'react';
import { renderHook, act } from '@testing-library/react-hooks';
import { wait } from '@testing-library/react';

test('should increment counter', async () => {
  const { result } = renderHook(useCounter);

  act(result.current.increment);

  expect(result.current.count).toBe(1);

  await wait(() => expect(result.current.count).toBe(2));
});

function useCounter() {
  const [count, setCount] = useState(0);

  useEffect(() => {
    setTimeout(() => {
      setCount(count => count + 1);
    });
  }, []);

  const increment = useCallback(() => setCount(x => x + 1), []);

  return { count, increment };
}
```

### What you did:

Tried to augment the example test in the README with an asynchronous effect.

### What happened:

Although the test passes I get a warning about using differing versions of `act()`:

```
 PASS  test/foo.test.tsx
  ✓ should increment counter (13ms)

  console.error node_modules/react-test-renderer/cjs/react-test-renderer.development.js:104
    Warning: It looks like you're using the wrong act() around your test interactions.
    Be sure to use the matching version of act() corresponding to your renderer:

    // for react-dom:
    import {act} from 'react-dom/test-utils';
    //...
    act(() => ...);

    // for react-test-renderer:
    import TestRenderer from 'react-test-renderer';
    const {act} = TestRenderer;
    //...
    act(() => ...);
        in TestHook
        in Suspense
```

mpeyper:
**What**:

Added additional async utils:
- `wait(callback)`
- `waitForValueToChange(selector)`

Resolves #173

**Why**:

These additional utils privide similar to the [`wait`](https://testing-library.com/docs/dom-testing-library/api-async#wait) and [`waitForElement`](https://testing-library.com/docs/dom-testing-library/api-async#waitforelement)/[`waitForElementToBeRemoved`](https://testing-library.com/docs/dom-testing-library/api-async#waitforelementtoberemoved) from the [`dom-testing-library` async utils](https://testing-library.com/docs/dom-testing-library/api-async).

This compliments the `waitForNextUpdate` utility we already provide that shares similarities to the [`waitForDomChange`](https://testing-library.com/docs/dom-testing-library/api-async#waitfordomchange) utility.

**How**:

Theses utilities are layers on top of `waitForNextUpdate`, running their relevant checks on each rerender of the test component.

I chose to be constsent with `dom-testing-library` by not pushing any values into the callbacks, but rather allow/expect the existing APIs to be used within them, this means testing for a value to change looks like:

```js
test('should wait for value to change', async () => {
  const { result, waitForValueToChange } = renderHook(() => useSomeHook())

  await waitForValueToChange(() => result.current.someValue === 'expected value')

  // perform assertions
}
```

instead of

```js
test('should wait for value to change', async () => {
  const { waitForValueToChange } = renderHook(() => useSomeHook())

  await waitForValueToChange((currentResult) => currentResult.someValue === 'expected value')

  // perform assertions
}
```

or something similar.

I anticipate this to cause some confusion when people first start using it so I have tried to call it out explicitly in the API reference docs.  I have not made any changes to the usage docs yet as I want to get a feeling on what people are struggling with and how best to describe the intent and usage before making those changes.

I'm not super sold on the name of `waitForValueToChange` and may change it, so if you have any suggestions, I'd love to hear them.

I'm also thinking about refactoring the implementation to have `wait` as the underlying utility instead of `waitForNextUpdate` by proving a `() => {}` callback to `wait`.  I think this might make more conceptual sense to anyone else coming in and trying to understand how these all fir together.

**Checklist**:

- [x] Documentation updated
- [x] Tests
- [ ] Ready to be merged


mpeyper:
> I'm also thinking about refactoring the implementation to have wait as the underlying utility

I had a stab at this and it was actually way more complicated than I expected to flip the `waitForNextUpdate` and `wait` relationship around due to how when the render listeners resolve.  This could probably be overcome with an overhaul of that system, but I chose not to do that at this time.

I did manage to refactor `waitForValueToChange` to use `wait` under the hood indead of duplicating much of the logic.  This introduced a new option into the mix, `suppressErrors`, which is `true` by default for `wait` but `false` by default for `waitForValueToChange`.  I'm considering not surfacing this as an API option but I'm not sure yet.

I've also been thinking about the name `waitForValueToChange` and I'm wondering if `waitForNextValue` is perhaps a bit better?  It's consistent with the `waitForNextUpdate` naming and a bit more concise to type, but I'm not sure it's as descriptive about what it's waiting for.  I dunno. Happy to hear thoughts on this.

mpeyper:
I've dragged my heels on this long enough on this.  I've decided to take the names as they are now and we can always change them if better suggestions come along or if people don't understand them.

PR #208, opened 2019-10-26
Include missing "dont-cleanup-after-each" script in npm package

ndresx:
**What**:

This adds the currently missing `dont-cleanup-after-each.js` file reference in `package.json` so that npm can pick it up and ship it within an upcoming release.

**Why**:

According to the docs (https://react-hooks-testing-library.com/reference/api#cleanup), it should be possible to set up this script via Jest config, but the file doesn't exist in the current package version (https://www.unpkg.com/browse/@testing-library/react-hooks@3.1.0/).

It also removes the redundant `README.md` and `LICENSE.md` from the `files` list as they are always included on `npm publish` (https://docs.npmjs.com/files/package.json#files).

**Checklist**:

- [x] Ready to be merged


mpeyper:
Wow. Whoops!

Thanks for spotting that and fixing it!

Issue #283, opened 2020-01-24
Call to function that changes hook state not wrapped in `act()` in documentation example


aryyya:
In the ["Advanced Hooks" documentation page, the Async section](https://github.com/testing-library/react-hooks-testing-library/blob/master/docs/usage/advanced-hooks.md#async) includes an example with a call `result.current.incrementAsync()` that is not wrapped in `act()` like all the other examples on the page are:

```js
import { renderHook } from '@testing-library/react-hooks'
import { useCounter } from './counter'

test('should increment counter after delay', async () => {
  const { result, waitForNextUpdate } = renderHook(() => useCounter())

  result.current.incrementAsync() // not wrapped in act()

  await waitForNextUpdate()

  expect(result.current.count).toBe(1)
})
```

If this example code is correct and the call is **not** supposed to be wrapped in `act()`, could there be an explanation as to why it differs from the synchronous examples?

mpeyper:
The example is correct. The update does not occur when the call is made, but asynchronously when the `await waitForNextUpdate()` is waiting. The async utils automatically wrap the waiting code in the asynchronous `act` wrapper so you don't need to.

I'm on vacation at the moment, but happy to review a PR modifying the docs to make this more clear.

For what it's worth, it shouldn't hurt to wrap the call in `act` as well, if you prefer a more uniform approach in you tests.

aryyya:
Thanks for the explanation! I created pull request #284 with the information that you gave me.

PR #284, opened 2020-01-24
Add explanation for async state updates not needing to be wrapped in …

aryyya:
…act()

**What**:

#283 Call to function that changes hook state not wrapped in `act()` in documentation example.

**Why**:

To clarify the difference in the required usage of `act()` between synchronous and non-synchronous test code.

**How**:

Explanatory text is added to the relevant documentation page.

**Checklist**:

- [x] Documentation updated
- [x] Ready to be merged

### DIFF
--- a/docs/usage/advanced-hooks.md
+++ b/docs/usage/advanced-hooks.md
@@ -132,6 +132,10 @@ test('should increment counter after delay', async () => {
 })
 ```
 
+Wrapping `incrementAsync` in `act()` is not necessary since the state updates happen
+asynchronously during `await waitForNextUpdate()`. The async utils automatically wrap the waiting
+code in the asynchronous `act()` wrapper.
+
 For more details on the the other async utilities, please refer to the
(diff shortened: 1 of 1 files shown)


docs/usage/advanced-hooks.md
```diff
@@ -132,6 +132,10 @@ test('should increment counter after delay', async () => {
 })
 ```
 
+Wrapping `incrementAsync` in `act()` is not necessary since the state updates happen
+asynchronously during `await waitForNextUpdate()`. The async utils automatically wrap the waiting
```
mpeyper: Can you please change "utils" to "utilities"? I try to avoid abbreviations in the docs to assist beginners and non-english speakers that might not be familiar with the terms (googling the real word will at least give them a definition).

aryyya: Not using abbreviations is a good idea. Fixed!

aryyya:
Fixed the issue, no need to add myself as a contributor. Thanks!

PR #304, opened 2020-03-01
docs: fix fragment identifier

numb86:
**What**:

Fix documentation.

**Why**:

Fragment identifier is incorrect.

**How**:

**Checklist**:

- [x] Documentation updated
- [x] Ready to be merged

### DIFF
--- a/docs/usage/advanced-hooks.md
+++ b/docs/usage/advanced-hooks.md
@@ -142,7 +142,7 @@ For more details on the the other async utilities, please refer to the
 ### Suspense
 
 All the [async utilities](/reference/api#async-utilities) will also wait for hooks that suspends
-using [React's `Suspense`](https://reactjs.org/docs/code-splitting.html#reactlazy) functionality to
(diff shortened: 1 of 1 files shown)


docs/usage/advanced-hooks.md
```diff
@@ -142,7 +142,7 @@ For more details on the the other async utilities, please refer to the
 ### Suspense
 
 All the [async utilities](/reference/api#async-utilities) will also wait for hooks that suspends
-using [React's `Suspense`](https://reactjs.org/docs/code-splitting.html#suspense) functionality to
+using [React's `Suspense`](https://reactjs.org/docs/code-splitting.html#reactlazy) functionality to
```
mpeyper: It was nice having the longer explanation for suspense, but I don't think linking to `React.lazy` is the best thing here (it's not going to be very common that hooks will be suspending for lazy components).

Perhaps the `Suspense` API reference will do instead?

https://reactjs.org/docs/react-api.html#reactsuspense

numb86: I agree. Fixed.

PR #385, opened 2020-06-07
Point-free examples

flextremedev:
**What**:

Refactoring examples to use point-free definitions.

**Why**:

Many users tend to refer to the example and it might be useful for the example to be as clean as possible as in this case free of redundancy.

**How**:

Deleted wrapping functions and only left the function definition point-free.

**Checklist**:

- [x] Documentation updated
- [x] Ready to be merged
      
This is my first open source contribution. Would be great to hear if I could do or could've done something better.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -174,6 +174,16 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "flextremedev",
+      "name": "Pascal Wegner",
+      "avatar_url": "https://avatars0.githubusercontent.com/u/16837852?v=4",
+      "profile": "https://github.com/flextremedev",
+      "contributions": [
+        "doc",
+        "example"
+      ]
     }
   ],
   "commitConvention": "none"
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ import { renderHook, act } from '@testing-library/react-hooks'
 import useCounter from './useCounter'
 
 test('should increment counter', () => {
-  const { result } = renderHook(() => useCounter())
+  const { result } = renderHook(useCounter)
 
   act(() => {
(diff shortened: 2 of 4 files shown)


mpeyper:
Hi @flextremedev,

Thanks for take the time to put this together and congrats on your first OOS contribution! May it be the first of many :)

The use of the wrapping function in our docs and in general is actually quite deliberate.  One of our goals is that the use of the hooks should be as natural and normal as possible to using them within a component, so we recommend and teach writing the hook callback for `renderHook` in full, even if there is technically some redundancy in doing so.

I can also say that in my experience, inexperienced developers can also find the expanded form easier to follow and find passing functions around to be confusing and uncomfortable at first (yes, I know they are still passing a function, but it feels different to them).

Finally, I have found that in practice, it does not take very long for [a custom hook to require arguments to passed to it](https://react-hooks-testing-library.com/usage/basic-hooks#providing-props), resulting in the shorthand needing to be expanded anyway.

Because of these reasons, I unfortunately won't be accepting this PR as I don't feel the point-free style aids readability or usability more than the benefits I mentioned above.  

> This is my first open source contribution. Would be great to hear if I could do or could've done something better.

I'm sorry your first contribution did not result in a merge today.  I would hate if that discouraged you from making future contributions here or in other projects.  Your approach was fine and it you did everything right, the change was just not in line with my vision of how the library should be used, that's all.

Usually I would recommend opening an issue and discussing the change instead of going straight to the PR as it can often save you some time, but given the size and scope of proposed changes here, this was likely just as easy to do.

Note: there are some failing checks on the PR for generating the docs.  I do not believe this has anything to do with you, so don't panic about that.  I'll be investing those failure separately.

Issue #68, opened 2019-05-13
Server side rendering


simmo:
### Describe the feature you'd like:

Support for server side rendering. I have rough implementation in the gist below... me than happy to drop this in a PR.

### Suggested implementation:

https://gist.github.com/simmo/9b8d9fb13b5fba513a76a8413eeeb805

### Describe alternatives you've considered:

I haven't seen any... yet.

### Teachability, Documentation, Adoption, Migration Strategy:

I guess an additional export? Happy to take some guidance! 😄

mpeyper:
Looks cool.

I wonder how this would feel as an option to `renderHook` instead of a seperate function? I imagine it's only the initial render that would use the server renderer and subsequent `rerender` calls would use the existing renderer (with rehydration or something)? I'm quite unfamiliar with server-side rendering, so this might be nonsense.

My other thought is that we recently changed to use `react-test-renderer` so that we don't need a DOM for testing the hooks. Does this get undone by `renderToString` from `react-dom` or does the fact it's the server mean it's still not required?

simmo:
Hey @mpeyper 

I'm thinking out-loud here... I was wondering if we could flag `react-dom` as an optional dependency as, like you mention, SSR might not be the no.1 use case.

And, yes, you raise a valid point about hydration. For example you'd maybe expect a `useInterval` hook ([example](https://overreacted.io/making-setinterval-declarative-with-react-hooks/)) to return an initial state and then hydrate on the client, triggering `useEffect`, executing the `setTimeout`.

I'll have a play with `react-test-renderer` and see what I can come up with. A bit of a head-scratcher!!! 🤔

mpeyper:
Hey @simmo,

just wondering how you're going with this and if there is anything I can help you with, or if this idea is dead and I can close this issue?

mpeyper:
Closing this due to inactivity. Happy to reopen if anyone wants to start the discussion back up.

fabb:
It would be really helpful to have this. I have a hook that produces different values on SSR and CSR based on useEffect and media queries. It would be great if there was a way to unit test this.

mpeyper:
I'd love some help with this as SSR is completely foreign to me.  I don't even understand what is failing/missing in the current implementation to begin looking at this.

I'll reopen this now as I'd also like to see us support more use cases.

fabb:
On SSR, `useEffect` hooks are not run.

mpeyper:
[...]

PR #387, opened 2020-06-08
Server Rendering

mpeyper:
**What**:

This is an early look at what I'm thinking for #68.

It adds the ability to test hooks in different environments by allowing different renderers to be used.  The library will ship with 3 renderers:

- `@testing-library/react-hooks/native` (`react-test-renderer`)
- `@testing-library/react-hooks/dom` (`react-dom`)
- `@testing-library/react-hooks/server` (`react-dom/server`)

Additionally, the main import for the package (`@testing-library/react-hooks`) will autodetect which of `react-dom` or `react-test-renderer` is installed and select the appropriate renderer to use.

**Why**:

Currently there was no way to test how a hook would behave when using rendering server side rendering.

#418 has also raised the issue that `refs` are not being set when using `react-test-renderer` so this would also see a solution for users requiring that but being able to use `react-dom` to render instead.

**How**:

By abstracting the rendering, rerendering and unmounting behaviour away from the core functionality, it was possible to provide thin renderer implementations that can leverage the same underlying logic.  This opened the door to allowing DOM users to use use the existing renderer they have installed instead of having to install `react-test-renderer` specifically for this library.

I would like to make it easier for other libraries to implement their own renderer on top of our core logic as well to make more use cases easier to implement (although I don't think we will ever ship more than the 3 renderers in this package).   This would likely be using a `@testing-library/react-hooks/custom` import, rather than drilling into the `lib/core` module.

Docs are still to come and tests are not complete (but pretty close).

**Checklist**:

- [ ] Documentation updated
- [ ] Tests
- [ ] Ready to be merged
      

mpeyper:
I'm going to close this as it's getting very far behind master and there hasn't been much interest raised in it.   If anyone is keen for it, or functionality like it, comment here and we might pick it up again.

PR #396, opened 2020-06-20
Link to new Discord

nickmccurdy:


mpeyper:
Thanks @nickmccurdy.

Can I get a `react-hooks` channel, or do you think we'd be more effective directing questions and discussions to the  regular `react` channel?

nickmccurdy:
I guess it should be separate since it has different goals and its own API. Created it.

PR #414, opened 2020-07-23
Upgrade the typings dependency and update the docs

erozak:
**What**:
- Upgrade the typings dependency
- Update the docs about the v3.3.0 changes of the `wrapper` option

**Why**:

- The typings dependency doesn't match the APIs

**Checklist**:

- [x] Documentation updated
- [x] Ready to be merged

### DIFF
--- a/docs/usage/advanced-hooks.md
+++ b/docs/usage/advanced-hooks.md
@@ -40,35 +40,19 @@ import { renderHook, act } from '@testing-library/react-hooks'
 import { CounterStepProvider, useCounter } from './counter'
 
 test('should use custom step when incrementing', () => {
-  const wrapper = ({ children, step }) => (
-    <CounterStepProvider step={step}>{children}</CounterStepProvider>
-  )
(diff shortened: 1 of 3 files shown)


docs/usage/advanced-hooks.md
```diff
@@ -40,19 +40,35 @@ import { renderHook, act } from '@testing-library/react-hooks'
...
+    }
+  })
 
   act(() => {
     result.current.increment()
   })
 
   expect(result.current.count).toBe(2)
+
+  rerender({ step: -2 })
```
mpeyper, line 59: Given that this is a relatively niche feature, I think we would be better off have the basic example just with a `wrapper` and then some text about using `props` with the `wrapper` before showing the more complex example.

Thoughts?

erozak: I moved the revision as a new section "Providing Props". It may help to show the use case about using `initialProps` and the `rerender` method with `wrapper`.

PR #448, opened 2020-09-10
Remove package lock

nickmccurdy:


mpeyper:
I don't necessarily disagree with this change, but is there any context around suggesting it?

nickmccurdy:
I'm migrating the organization to not use package locks for published packages for consistency with existing packages

PR #453, opened 2020-09-18
Update `@types/testing-library__react-hooks` dependency version

cliffzhaobupt:
**What**:

Update the version of dependency `@types/testing-library__react-hooks`

**Why**:

In 3.3.0, there's no type definition of `waitFor`, so in my project I have to use the deprecated `wait` instead.

**How**:

Run `npm update @types/testing-library__react-hooks`

**Checklist**:

- [x] Documentation updated
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3696,9 +3696,9 @@
       }
     },
     "@types/react-test-renderer": {
-      "version": "16.9.2",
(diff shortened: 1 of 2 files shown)


mpeyper:
I wonder if we can configure @renovate-bot to keep that one up to date, even though it's not pinned, without including all the regular dependencies?  I'll look into that.

Thanks 😃

PR #471, opened 2020-10-15
pr/docs-fix-links

gortron:
**What**:

Some links in the documentation were broken, giving user's browsing the documentation on GitHub a 404 error.

The links are working fine as is on your [documentation site](https://react-hooks-testing-library.com/), they are only broken on GitHub. It's possible these fixes will causes issues there. If that's the case, please feel free to reject and close this PR.

**Why**:

Fixing the links makes it easier for people to browse the documentation.

**How**:

The links are fixed by updating their routes where necessary. 

**Checklist**:

- [ x ] Documentation updated
- [ x ] Added myself to contributors table

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -183,6 +183,15 @@
       "contributions": [
         "test"
       ]
+    },
+    {
+      "login": "gortron",
+      "name": "Gordy Lanza",
+      "avatar_url": "https://avatars3.githubusercontent.com/u/6877497?v=4",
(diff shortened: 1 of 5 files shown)


mpeyper:
Apologies for taking so long to get to this.

The docs are intended to be used in the static site and not directly in github and I think you are right that these changes will mess with the built version.

Docz has been giving me all kinds of issues recently so I'm thinking about replacing the whole lot anyway.

Thanks for the time effort, but I'm just going to leave it as is for now.

PR #496, opened 2020-12-03
Capture all results a rendered hook produces.

### LINKED ISSUE #461: Capture intermediate updates when hook rerenders
### What is your question:

Let's consider the following hook
```jsx
const useCounter = () => {
    const [counter, setCounter] = React.useState(0);

    React.useEffect(() => {
        setCounter(1)
    }, [])

    return counter;
}

export default useCounter;
```

when running a test using RHTL, the following passes:
```js
describe('useCounter', () => {
   it('should  provide the initial count', () => {
       const { result } = renderHook(() => useCounter());
       expect(result.current).toBe(1); // ----> this passes 🟢
   });
});
```

But this test does not capture the actual behavior of how a component consumes a hook during different render cycles
```jsx
function App() {
  const counter = useCounter();
  console.log('counter', counter);
 return counter; // returns 0 during first render and then returns 1 during second render.
}
```

Is there a way to test the output of the first render using RHTL? I'm currently in a situation where I just fixed a bug related to variable initialization in a custom hook and it would be nice to be able to add a test to increase my code coverage.

mpeyper:
Fixes #461

**What**:

Make all results a rendered hook produces available to tests for assertions.

**Why**:

To test that initial states (prior to effects running) or other intermediate states were produced as expected.  There has been some suggestion this could be useful for testing SSR scenarios as well.

**How**:

When the result gets update it we push it into an array and surface it through the `result` container returned from `renderHook` as `result.all`.

**Checklist**:

- [x] Documentation updated
- [x] Tests
- [x] Ready to be merged
      

mpeyper:
I want to hold off on this until after #502 so I can test semantic release (I'll amend my commit messages first).

Issue #498, opened 2020-12-04
Convert to TypeScript


mpeyper:
### Describe the feature you'd like:

Title says it all.  I feel like TypeScript has reached a level of adoption where we shouldn't be scared of losing contributors by using it instead of JavaScript.

I believe other `@testing-library` packages are considering or currently being converted, if they aren't already TS.

### Suggested implementation:

We currently have [type definitions in `DefinitelyTyped`](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/testing-library__react-hooks) which can be used as the basis of the types.

I think we can aim to have a straight cut over to a new codebase rather than a progressive migration as there are not that many files to consider.

I think we should look at using [TSDX](https://tsdx.io/) for the tooling unless anyone has other ideas or suggestions in this space.

### Describe alternatives you've considered:

N/A

### Teachability, Documentation, Adoption, Migration Strategy:

This package should also become the source of truth for the type definitions and the `DefinitelyTyped` version should be deprecated.  I'm not sure what the process if to do that if there is one.

Other than that I don't think there are any particular considerations to be made for this.

joshuaellis:
The compiler won't recommend looking for `@types/*` library anymore because we'll have bundled types. So I imagine they would be able to put a depreciated flag on the package?

Is there a reason you think we should `tsdx`? Looking at the repo, it's pretty immature and there's a few outstanding issues. I think we could probably just use our own configuration. Proabably easier to maintain overall.

nickmccurdy:
> I think we should look at using [TSDX](https://tsdx.io/) for the tooling unless anyone has other ideas or suggestions in this space.

`kcd-scripts` now supports TypeScript and it's already the de facto standard tool for this organization.


> This package should also become the source of truth for the type definitions and the `DefinitelyTyped` version should be deprecated. I'm not sure what the process if to do that if there is one.

There's a list of removed packages that you can submit a pull request to.

mpeyper:
> Is there a reason you think we should `tsdx`?

Only that I have seen folks on Twitter mention how nice it was to use.

I'm also happy with kcd-scripts if that is the preferred approach.

marcosvega91:
If you want I can migrate the actual codebase to kcd-scripts.

nickmccurdy:
In that case it's probably easier to do #499 with [`yo kcd-oss`]() first. We'll have better CI for this PR and TS tooling setup will be simpler.

[...]

PR #510, opened 2020-12-06
feature/RHTL-68 – server-side-rendering

joshuaellis:
**What**:

New feature SSR testing – breaking change.
Split the repo into `dom|native|server`

```
dom: react-dom
native: react-test-renderer
server: react-dom
```

**Why**:

#68

**How**:

RTR does not currently support SSR, `react-dom` does, this will be the default renderer for SSR testing. However, the plan is to expose a `custom` renderer usage where users can use other renderers e.g `preact`. User's will have their test engine auto

**Checklist**:

- [ ] Documentation updated
- [x] Tests
- [ ] Ready to be merged
      

joshuaellis:
This isn't ready but i've made the move over into 'native' and thought it'd be good to track along side #498 as that is also a huge change to the codebase.

joshuaellis:
the `.eslintrc` config is not being picked up hence the lint errors. Also I can't actually run `validate` or `lint` because of the following error
```shell
Error: .eslintrc » ./node_modules/kcd-scripts/eslint.js » /Users/joshua/code/react-hooks-testing-library/node_modules/kcd-scripts/node_modules/eslint-config-kentcdodds/jest.js:
        Environment key "jest/globals" is unknown
```
My assumption this is related to #502 any ideas @marcosvega91?

marcosvega91:
Hi @joshuaellis have you reinstalled all modules ? Are you using `yarn` or `npm`?.

If you are using `yarn` remember to remove your local `yarn.lock` then Delete `node_modules` folder and install all libraries again. Let me know if this solve your problem

joshuaellis:
Hi @marcosvega91,

>If you are using yarn remember to remove your local yarn.lock then Delete node_modules folder and install all libraries again.

I was using `yarn`. But doing the above method gave the same error. repeating steps above and re-installing with `npm` worked fine. Does `kcd-scripts` not support yarn? Weird if not.

joshuaellis:
Hi @marcosvega91,

>If you are using yarn remember to remove your local yarn.lock then Delete node_modules folder and install all libraries again.

I was using `yarn`. But doing the above method gave the same error. repeating steps above and re-installing with `npm` worked fine. Does `kcd-scripts` not support yarn? Weird if not.

marcosvega91:
You're right. I'm receiving the same error. I'll try to understand why with yarn is giving this problem. Thanks for raising it 😄

marcosvega91:
The problem with yarn is that is not resolving correctly `eslint`. `yarn` is using `6.8` will it should use the version `7`

joshuaellis:
Hi @mpeyper,

I've just migrated the previous server work you did & the `server/suspenseHook` fails because - 
```shell
ReactDOMServer does not yet support Suspense.
      31 |       renderProps = props
      32 |       baseAct(() => {
    > 33 |         const serverOutput = ReactDOMServer.renderToString(toRender(props))
         |                                             ^
      34 |         container.innerHTML = serverOutput
      35 |       })
      36 |     },

```
I just wanted to know if you remember this happening to you or not? IMO i can't initially see why this test would fail as we're not using `suspense`.  Maybe i'm not understanding the error well enough. If you're not sure dw.

mpeyper:
@joshuaellis Yes, this was an issue for me too.

mpeyper:
@joshuaellis the TS migration (#498, #520) is in a beta release now.  I suggest we rebase these changes on the [`beta` branch(https://github.com/testing-library/react-hooks-testing-library/tree/beta) before continuing.

joshuaellis:
Great stuff @mpeyper thanks for letting me know.

joshuaellis:
I've managed to navigate alot of the Typescript stuff so far, I'm really struggling how to solve the issue in `native/pure` with the `toRender` –
```tsx
const toRender = (props?: TProps): JSX.Element => (
    <Wrapper {...props}>
      <TestHook hookProps={props} {...testHookProps} />
    </Wrapper>
  )
```

the issue i'm getting is
> Type '{ children: Element; }' is not assignable to type 'TProps'.
  'TProps' could be instantiated with an arbitrary type which could be unrelated to '{ children: Element; }'.

I was wondering if it had something to do with this on going issue – https://github.com/DefinitelyTyped/DefinitelyTyped/issues/34237 if anyone has ideas, they're definitely welcome.

joshuaellis:
> I've managed to navigate alot of the Typescript stuff so far, I'm really struggling how to solve the issue in `native/pure` with the `toRender` –
> 
> ```tsx
> const toRender = (props?: TProps): JSX.Element => (
>     <Wrapper {...props}>
>       <TestHook hookProps={props} {...testHookProps} />
>     </Wrapper>
>   )
> ```
> 
> the issue i'm getting is
> 
> > Type '{ children: Element; }' is not assignable to type 'TProps'.
> > 'TProps' could be instantiated with an arbitrary type which could be unrelated to '{ children: Element; }'.
> 
> I was wondering if it had something to do with this on going issue – [DefinitelyTyped/DefinitelyTyped#34237](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/34237) if anyone has ideas, they're definitely welcome.

DM, it actually had to do with this issue https://github.com/Microsoft/TypeScript/issues/28938#issuecomment-450636046

mpeyper:
I updated the base branch to reduce the noise in the review and I think we'll want this to honour as a beta first anyway.

I've had a brief look on my phone just now and I like where this is headed. I have a few comments to leave, but I'll wait until I'm on my laptop to do so.

src/types.ts
```diff
@@ -0,0 +1,139 @@
+import { act as RTRAct } from 'react-test-renderer'
```
mpeyper, line 1: I'm not sure we can have this import here as we can rely on them having this dependency installed and TS being statically analysed, it's going to straight up fail for consumers without it.

It sucks, but we might need to just define our own interface for `act` and let the duck typing take care of it. Thoughts?

joshuaellis: Yeah I wasn't sure about this. We could make our own wrapper similar to how it looks in `server/pure` –

```ts
export type ServerModifiedAct = (
  cb: ServerActCallbackAsync | ServerActCallback
) => Promise<undefined> | void
```

I'm also concerned how the type will work for `createRenderHook` in `core/index` when it's built whether it will correctly assert if you're using `react-hooks-testing-library/server` you **will** get hydrate in the return. Probably something to test in beta.

joshuaellis:
I think this is good to review fully now, I don't believe i've missed anything. When we're happy with the implementation I'll branch again off this to work on the documentation so there's less noise when we're reviewing that.

package.json
```diff
@@ -40,6 +42,7 @@
   "dependencies": {
     "@babel/runtime": "^7.12.5",
     "@types/react": ">=16.9.0",
+    "@types/react-dom": "^17.0.0",
```
mpeyper, line 45: This should be `>=16.9.0` as well

package.json
```diff
@@ -14,6 +14,8 @@
   "files": [
     "lib",
     "src",
+    "native",
+    "server",
```
mpeyper, line 18: This should also include `dom`

src/core/cleanup.ts
```diff
@@ -16,4 +16,13 @@ function removeCleanup(callback: () => Promise<void> | void) {
   cleanupCallbacks = cleanupCallbacks.filter((cb) => cb !== callback)
 }
 
+cleanup.autoRegister = () => {
```
mpeyper, line 19: I'm second guessing myself on whether it's a good idea to put this on `cleanup` or to just have it as another export.  I think currently it will be exposed out of the public API, which might be ok, but then we should to document and maintain it.

mpeyper: Don't change this yet... I'll sleep on it.

src/pure.ts
```diff
@@ -0,0 +1,31 @@
+import { RenderingEngineArray, ReactHooksRenderer } from 'types'
+
+const RENDERERS: RenderingEngineArray = [
+  { required: 'react-test-renderer', renderer: './native/pure' },
+  { required: 'react-dom', renderer: './server/pure' }
```
mpeyper, line 5: The default renderer for `react-dom` should be `./dom/pure`.  Server rendering is opt in for testing edge cases.

src/types.ts
```diff
@@ -0,0 +1,148 @@
...
+ *
+ * dom/pure
+ *
+ */
+
+export interface DomRendererReturn<TProps> extends GenericRendererReturn<TProps> {
+  act: ReactDomAct
+}
+
+export type DomRendererOptions<TProps> = GenericRendererOptions<TProps>
```
mpeyper, line 148: Could all the `...RendererOptions` be refactored into a single `RendererOptions` type?  I don't think there is a need for 3 additional aliases, unless I'm missing something?

src/core/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+function createRenderHook(
+  createRenderer: <TProps, TResult>(
+    testProps: Omit<TestHookProps<TProps, TResult>, 'hookProps'>,
+    opts: NativeRendererOptions<TProps>
+  ) => NativeRendererReturn<TProps> | DomRendererReturn<TProps>
+): <TProps, TResult>(
+  callback: (props: TProps) => TResult,
+  opts?: RenderHookOptions<TProps>
+) => RenderHookReturn<TProps, TResult>
+function createRenderHook(
```
mpeyper, line 80: I've got some ideas to reduce the noise here and make `createRenderHook` agnostic of any reference to the type of renderer.  It's a bit difficult to explain, so I might put up a PR to your branch with what I'm thinking tomorrow.

src/types.ts
```diff
@@ -0,0 +1,148 @@
+/**
```
mpeyper, line 1: Overall, I think this file has too many types in it.  We already have the modules for the different renderers so I think we should move the specific types into those which import shared types from here (or `core`?).

Personally, I prefer keeping the types with code, but I haven't worked on a TS project with as much crossover in types between modules before so this approach might be better?

As mentioned above, I've got some ideas to reduce the amount of Renderer specific stuff required so I'll see after I propose those changes if there is any more cleanup here.

test/dom/asyncHook.ts
```diff
@@ -1,5 +1,5 @@
 import { useState, useRef, useEffect } from 'react'
-import { renderHook } from '../src'
+import { renderHook } from '../../src/dom'
```
mpeyper, line 2: I feel like I'm getting very close to moving the tests into `src/**__tests__/*` instead of the seperate root they're in now.

src/native/pure.ts
```diff
@@ -0,0 +1,39 @@
...
+        container.unmount()
+      })
+    },
+    act
+  }
+}
+
+const renderHook = createRenderHook(createNativeRenderer)
+
+export { renderHook, act, cleanup, addCleanup, removeCleanup }
```
mpeyper, line 39: It looks like all the other renderers and the auto detection export are missing `addCleanup` and `removeCleanup` in their exports (they were added after my initial POC for this).

src/types.ts
joshuaellis: I agree, on reflection we can consolidate. I think I seperated without knowing if there was gonna be a outlier.

src/core/index.tsx
joshuaellis: That would be great, my main challenge was have TS understand when hydrate would exist and when it wouldn't, hence the overload.

src/types.ts
joshuaellis: My personal preference is this way, you can see & compose types really simply. But I'm open to doing it the alternative way.
>As mentioned above, I've got some ideas to reduce the amount of Renderer specific stuff required so I'll see after I propose those changes if there is any more cleanup here.

I agree, lets do it after the rest of the changes.

test/dom/asyncHook.ts
joshuaellis: My personal preference to have the tests as close to the code as possible, but I didn't want to change what you'd already done. If you're happy to do it that way, i'd prefer it.

src/native/pure.ts
joshuaellis: I can add it, i'm wondering what the use case is for exposing the addCleanup & removeCleanup API?

mpeyper: [There were reasons](https://github.com/testing-library/react-hooks-testing-library/pull/432#issuecomment-687766727)... Still not sure it was the right call though.

test/dom/asyncHook.ts
mpeyper: My preference for a seperate test root stems mostly from my Java background. I'm coming around to nested test directories and paths like this show the benefit.

This review is noisy enough with changes, so perhaps leave it for this PR and we can spin up a new one to move the tests once we're happy with the implementation (like you suggested for the docs)

src/core/index.tsx
mpeyper: The idea I had would infer extra functions from the result of `createRenderer` and the resulting type _just works™️_.  The function signature might get a bit crazy, but the usage won't change.  We'll see how it goes 😅

src/core/cleanup.ts
mpeyper: Still undecided. I'm leaning towards just another export.

joshuaellis:
Stil to do [this comment](https://github.com/joshuaellis/react-hooks-testing-library/pull/1#discussion_r552510412) & define some tests for ‘src/pure’

mpeyper:
> Stil to do this comment & define some tests for ‘src/pure’

I've fixed the types for the submodules now and I've removed the `custom` module (but left the generics as discussed in that comment).

I worked around the type issue by generating the submodules instead of having static files for them.  In some ways I like this solution more, but I'm still a bit upset I couldn't figure out the TS configuration to get it to work.

As for tests of the root module, I'm still not sure how to do this as we have both dependencies installed so testing the error cases or even getting it to select the `dom` renderer is going to be a challenge.  I'm not a fan of exporting functions just for testing, but it might be the only way for this one?  I'm also happy to put it on the back burner for now.  The logic is fairly straight forward and unlikely to change much in the future.

Other than that, I think it's just the docs that need some love on this one now.

src/core/asyncUtils.ts
```diff
@@ -1,28 +1,15 @@
...
 
-function asyncUtils(addResolver: (callback: () => void) => void) {
+function asyncUtils(act: Act, addResolver: (callback: () => void) => void): AsyncUtils {
   let nextUpdatePromise: Promise<void> | null = null
 
   const waitForNextUpdate = async ({ timeout }: Pick<WaitOptions, 'timeout'> = {}) => {
-    if (!nextUpdatePromise) {
+    if (nextUpdatePromise) {
+      await nextUpdatePromise
+    } else {
```
mpeyper, line 12: small refactor here.  in the old code, we were double `await`ing the same promise.  Minor issue with almost zero impact in practice, but made debugging the async test issues I saw when trying to implement a preact renderer more difficult than it should have been

src/helpers/createTestHarness.tsx
```diff
@@ -0,0 +1,42 @@
+import React, { Suspense } from 'react'
+
+import { RendererProps } from '../types'
+import { WrapperComponent } from '../types/react'
+
+import { isPromise } from './promises'
+
+function TestHook<TProps, TResult>({
```
mpeyper, line 8: When looking at the `preact-hooks-testing-library` code I noticed they had renamed this to `TestComponent`.  I think I like that better as that is what it is.  It isn't a hook.

I'm happy to leave it, but given the size of the refactoring, it might be appropriate to rename it now.

mpeyper: I just did it 😛

mpeyper:
I think we should release this into a beta now that the previous beta has been released.  What do you think @joshuaellis?

joshuaellis:
Absolutely, let’s push to beta. I can begin work on the docs later today 💪🏼

mpeyper:
hmmmm... I was expecting that to publish as `5.0.0-beta.1`.  The bot even picked up the `BREAKING CHANGES`  in the release tag?

joshuaellis:
bad bot. That is weird, I would have expected that too.

marcosvega91:
to trigger the release of the `5` beta I guess you should merge master into beta first and then merge the PR into the beta branch. 

Your master branch has a commit `Merge branch 'beta'` that is linked to the `4` version.

mpeyper:
Thanks @marcosvega91, I have rebuilt the `beta` branch no with the `4.0.0` tag in the history so the next build should be right 🤞

mpeyper:
I just tried the beta out in one of my own projects.  All renderers are working great and the auto-detection is also working.  The only issue so far is that when niether `react-dom` or `react-test-renderer` are installed, the error message is coming out as

> Error: Could not auto-detect a React renderer. Are you sure you've installed one of the following
>  - react-dom/n  - react-test-renderer

[Whoops](https://github.com/testing-library/react-hooks-testing-library/blob/beta/src/pure.ts#L27).  I'll fix that quickly and get a new build out.

PR #511, opened 2020-12-07
feat(cleanup): remove unnecessary flush microtasks as already handled by act

mpeyper:
**What**:

Remove some particularly complicated code that flushes microtasks as part of cleaning up the rendered hooks.

**Why**:

I noticed [this change in RTL](https://github.com/testing-library/react-testing-library/pull/768) from which we originally _borrowed_ the `flush-microtasks.js` implementation from.  The interesting part is that we already have our cleanup wrapped in `act` so the code was even less required.

**How**:

I deleted the file and removed the import and usage of it.  

One difference to the RTL PR to note is that they also change `cleanup` from an `async` function to a regular function.  I chose not to do this as we recently introduced the `addCleanup` functionality in [`v3.5.0`](https://github.com/testing-library/react-hooks-testing-library/releases/tag/v3.5.0) which allowed user provided cleanup to be `async` and I didn't want to restrict them to being sync only, plus it avoided a breaking change.

**Checklist**:

- [x] Tests
- [x] Ready to be merged
      

I wonder if this should go out into a beta release like they did in RTL or if their beta is proof that it's fine and we just release the minor?


joshuaellis:
When did they release this beta? Were there any issues open against it?

mpeyper:
[Beta was realeased on 30/8](https://github.com/testing-library/react-testing-library/releases/tag/v11.0.0-beta.1) and [full release went out on 2/9](https://github.com/testing-library/react-testing-library/releases/tag/v11.0.0).

I don't know if there were any issues raised, but there were no addition commits between bets and release.

joshuaellis:
@mpeyper Can't see any issues relating to that release open or closed & the comments in the PR seem really positive. IMO we could just release this.

mpeyper:
ok, here we go...

PR #509, opened 2020-12-06
Add Dependabot

Goldziher:
**What**:

Added a `dependabot.yml` config to the .github folder

**Why**:

Due to issue #500 

**How**:

Following the contribution guide

**Checklist**:

- [ - ] Documentation updated
- [ x ] Tests
- [ x ] Ready to be merged
      
- [ x ] Added myself to contributors table

PR #512, opened 2020-12-07
chore: add eslint as dep of the project

marcosvega91:
**What**: force install eslint 7

**Why**: to avoid problem with gatsby that is using an old version

**How**:

**Checklist**:

- [ ] Documentation updated
- [ ] Tests
- [X] Ready to be merged
      
### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "react": "17.0.1",
     "react-test-renderer": "17.0.1",
     "typescript": "4.1.2",
-    "eslint": "^7.14.0"
+    "eslint": "7.14.0"
   },
   "peerDependencies": {
     "react": ">=16.9.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -48,7 +48,8 @@
     "kcd-scripts": "7.5.1",
     "react": "17.0.1",
     "react-test-renderer": "17.0.1",
-    "typescript": "4.1.2"
+    "typescript": "4.1.2",
+    "eslint": "^7.14.0"
```
mpeyper, line 52: Renovate used to pin `devDependencies`, but I'm not sure what Dependabot expects since #509.

I don't particularly mind either way, but I'd like them to be consistent. Maybe we unpin them all now?

marcosvega91: yes right, I have removed the pin :)

PR #515, opened 2020-12-09
Convert to TypeScript

juhanakristian:
**What**:

Converting library to TypeScript, issue #498 
Removes deprecated `wait` util (as per discussion in KCD Discord)
https://discord.com/channels/715220730605731931/785649901782433852/786338242101379142

Creating this PR on behalf of @tigerabrodi we and other people from KCD Discord are going to work on it as a group

**Checklist**:

- [ ] Documentation updated
- [ ] Tests
- [ ] Ready to be merged
      

mpeyper:
Glad to have you. Feel free to ask any questions and push up any changes for a review before it's ready.

I look forward to seeing what you all come up with.

tigerabrodi:
> Glad to have you. Feel free to ask any questions and push up any changes for a review before it's ready.
> 
> I look forward to seeing what you all come up with.

@mpeyper Luckily Batman is here, in case an emergency happens :stuck_out_tongue_closed_eyes:

src/cleanup.ts
```diff
@@ -1,4 +1,4 @@
-let cleanupCallbacks = []
+let cleanupCallbacks: (() => Promise<void>)[] = []
```
merodiro, line 1: I think the return type of callbacks should be `Promise<void>|void` in case the callback is not async

nobrayner: I had done it this way, as the callbacks are always called with `await` - Does it still make sense to be `Promise<void> | void` still?

merodiro: `await` can still work if the code is synchronous you can see an example in [the documentation for the await operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await).
so it makes sense to use `await` in this case because it may or may not be a promise

tigerabrodi:
When handling errors in `asyncUtils`, at two places I need to check if timeout does exist in unknown, how can I go about doing this?

Screenshot from 2020-12-10 21-44-51

Do you have any ideas guys, @TkDodo & @eps1lon, your input on this?

src/cleanup.ts
nobrayner: @merodiro Thanks! Fixing that now...

JacobMGEvans (commented):
Progress is looking awesome!!

JacobMGEvans:
@kentcdodds @mpeyper I was hoping to get input on this. Currently, if I remove the `undefined` type which would only allow the user to `return` something in the `waitFor()`. The questions would be, how opinionated does the library want to be in how users employ the utility. I also currently have these types on the `T | Promise<T> | undefined | VoidFunction`
Screen Shot 2020-12-10 at 7 36 23 PM

JacobMGEvans:
> @kentcdodds @mpeyper I was hoping to get input on this. Currently, if I remove the `undefined` type which would only allow the user to `return` something in the `waitFor()`, there are tests that are structured in a way that would fail. The questions would be, how opinionated does the library want to be in how users employ the utility. I also currently have these types on the `T | Promise<T> | undefined | VoidFunction`
> Screen Shot 2020-12-10 at 7 36 23 PM
@merodiro noticed if we utilize the OR operator with the types we wanted to pass in the behavior & types we wanted would work with the current tests.

src/asyncUtils.ts
```diff
@@ -33,26 +40,32 @@ function asyncUtils(addResolver) {
...
         return callbackResult || callbackResult === undefined
-      } catch (e) {
+      } catch (error: unknown) {
         if (!suppressErrors) {
-          throw e
+          throw error as Error
         }
+        return undefined
       }
     }
```
JacobMGEvans, line 64: @kentcdodds @mpeyper This is the current Types for `waitFor()` the type is currently expecting "something" to be returned. We could make it more explicit to `never` expect `void` `null` etc...

mpeyper: There are two use cases for `waitFor` in this library (I'm not sure how it behaves in RTL)

1. `waitFor` this thing to not throw an exception:
   ```js
   waitFor(() => {
     expect(result.current).toBe(expectedValue)
   })
   ```
   In this case, the only value that would break the waiting is if the callback returns `undefined` (no return).
2. `waitFor` this thing to return `truthy`
   ```js
   waitFor(() => result.current.state === expectedValue)
   ```
   In this case, the wait will only break when the `true` (or any `thruthy` value) is returned.

Does that help?

JacobMGEvans (commented):
- Minimum types started for Pure file
- File needs better, improved typing
- Refactoring needed for ESLint Disables to be removed IF POSSIBLE

src/asyncUtils.ts
JacobMGEvans: Definitely helps! Thanks.

src/pure.tsx
```diff
@@ -70,22 +76,30 @@ function renderHook(callback, { initialProps, wrapper } = {}) {
...
     act(() => {
       update(toRender())
     })
   }
 
   function unmountHook() {
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
     act(() => {
       removeCleanup(unmountHook)
       unmount()
```
JacobMGEvans, line 105: The Pure file needs to be heavily reviewed and ensure that the types are ideal for user interfacing.

JacobMGEvans:
It's gonna pass... 😅  lol

src/pure.tsx
```diff
@@ -48,7 +52,7 @@ function resultContainer() {
     }
   }
 
-  const updateResult = (value: unknown, error?: unknown) => {
+  const updateResult = (value?: R, error?: Error) => {
```
JacobMGEvans, line 55: Nice!

JacobMGEvans:
previous commit:  All tests are strongly typed with minimal types
to allow for working and made sure tests types were easily usable
with types in Pure and Utils file, allow for good UX & DX

JacobMGEvans:
@mpeyper This is likely ready for review.

tigerabrodi:
Well done heroes! The villain is soon defeated :tada: 

Just gotta wait for Batman :grin: :raised_hands:

mpeyper:
Thanks so much for the effort!

I'll give this a review tonight, which is about 13 hours away in my timezone.

src/pure.tsx
```diff
@@ -0,0 +1,125 @@
...
+// TODO: Add better type, currently file is utilizing minimum types to work
+// TODO: Attempt to refactor code to remove ESLint disables if possible
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type Props<T = any, R = any> = {
+  callback: (props: T) => R
+  hookProps: unknown
+  onError: CallableFunction
+  children: CallableFunction
+}
+function TestHook({ callback, hookProps, onError, children }: Props) {
```
merodiro, line 15: I'm not sure if this needs to be generic but we are not calling passing anything to it so it will always be any

src/pure.tsx
```diff
@@ -0,0 +1,125 @@
...
+    addResolver: (resolver: VoidFunction) => {
+      resolvers.push(resolver)
+    },
+    setValue: (value: R) => updateResult(value),
+    setError: (error: Error) => updateResult(undefined, error)
+  }
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function renderHook<T = any, R = any>(
```
merodiro, line 71: I think we can remove the default value (`any`) because it will be inferred

src/pure.tsx
```diff
@@ -0,0 +1,125 @@
+import React, { ReactElement, ReactNode, Suspense } from 'react'
+import { act, create, ReactTestRenderer } from 'react-test-renderer'
+import asyncUtils from './asyncUtils'
+import { cleanup, addCleanup, removeCleanup } from './cleanup'
+
+// TODO: Add better type, currently file is utilizing minimum types to work
```
nobrayner, line 6: We probably don't need these TODO lines anymore?

src/pure.tsx
```diff
@@ -0,0 +1,125 @@
...
+// TODO: Attempt to refactor code to remove ESLint disables if possible
```
nobrayner, line 7: We probably don't need these TODO lines anymore?

src/pure.tsx
nobrayner: @merodiro It should be inferred from the callback wich is passed on line 69:
image

merodiro: `<TestHook>` isn't generic and isn't passing anything to `Props` so it will always be any

src/pure.tsx
JacobMGEvans: We do not, please remove them.

JacobMGEvans, line 7:
```suggestion
```
Seeing if I can inline a suggestion and commit

JacobMGEvans: The answer is yes lol

src/pure.tsx
nobrayner: @merodiro Oh 🤦‍♂️ I completely missed that

src/pure.tsx
mpeyper: Agreed that these should ideally be inferred in s many situations as possible.

I'm a bit worried about the scenario where `initialProps` is not provided in the `renderHooks` call, but new props are provided in the `rerender` call, e.g.

```ts
test('should infer from initialProps', () => {
  const { rerender } = renderHook(({ arg }) => useSomeHook(arg), {
    initialProps: { arg: false }
  })

  rerender({ arg: true })
})

test('should infer from defaulted args', () => {
  const { rerender } = renderHook(({ arg = false } = {}) => useSomeHook(arg))

  rerender({ arg: true })
})
```

(note: the inferring from the defaulted args is the ideal goal, but might not actually have enough information to infer  successfully, even with the best typings in the world)

src/pure.tsx
mpeyper: I think `TestHook` should be made generic.

src/pure.tsx
```diff
@@ -0,0 +1,123 @@
...
+import asyncUtils from './asyncUtils'
+import { cleanup, addCleanup, removeCleanup } from './cleanup'
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type Props<T = any, R = any> = {
+  callback: (props: T) => R
+  hookProps: unknown
+  onError: CallableFunction
+  children: CallableFunction
+}
```
mpeyper, line 12: I think `Props` would be better as

```ts
type TestHookProps<T, R> = {
  callback: (props: T) => R
  hookProps: R
  onError: (error: Error) => void
  children: (value: R) => void
}
```

src/pure.tsx
```diff
@@ -0,0 +1,123 @@
+import React, { ReactElement, ReactNode, Suspense } from 'react'
+import { act, create, ReactTestRenderer } from 'react-test-renderer'
+import asyncUtils from './asyncUtils'
+import { cleanup, addCleanup, removeCleanup } from './cleanup'
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type Props<T = any, R = any> = {
```
mpeyper, line 7: In general, I prefer generic types to better describe what they are genericizing, e.g. `TProps` and `TResult` make following their usage (and intellisense) much easier to me.

mpeyper (commented):
I haven't reviewed `asyncUtils` yet, but ran out of time tonight.  Other than the comments I've left a few other notes:

1. running `npm run build` produces types into nested directories in the `lib` directory.  This is because of the seperated `test` root.  I'm happy if you want to move them into `src/__tests__` to make the issue go away.
2. you haven't defined a `types` field in `package.json`.  This is required if the generated types are no at `/index.d.ts` of the repo's root (`kcd-scripts` does not output them there).  I think `lib/index.d.ts` will work if you do `1` of this list
3. There's a lot of eslint rules being disabled.  Generally I'm against doing this on a per-line basis.  either the rule is bogus for this project (like the dangling promise one and `act`) and should be turned off globally, or the issue should actually be addressed.

I'll continue with this tomorrow.

tigerabrodi:
@mpeyper Thanks Batman :smiling_face_with_three_hearts:, will continue fighting the villains :smile:, I promise you :triumph:! :muscle: 

I will take a look at this now and try to do what I can, I will also take a second look at `asyncUtils`, perhaps I find something in there, that could be bettered :tada:

test/suspenseHook.ts
```diff
@@ -1,10 +1,10 @@
 import { renderHook } from '../src'
 
 describe('suspense hook tests', () => {
-  const cache = {}
-  const fetchName = (isSuccessful) => {
+  const cache: { value?: Promise<string> | string | Error } = {}
```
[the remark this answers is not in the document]
tigerabrodi: Has been fixed :tada:

src/cleanup.ts
```diff
@@ -7,12 +7,12 @@ async function cleanup() {
   cleanupCallbacks = []
 }
 
-function addCleanup(callback) {
+function addCleanup(callback: () => Promise<void> | void) {
   cleanupCallbacks.unshift(callback)
```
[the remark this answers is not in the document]
tigerabrodi: Seems like a good one, updated :tada:

src/asyncUtils.ts
```diff
@@ -1,28 +1,35 @@
...
+  interval?: number
+  timeout?: number
+  suppressErrors?: boolean
+}
+
+class TimeoutError extends Error {
+  timeout = true
+}
+
+function createTimeoutError(utilName: string, { timeout }: Pick<WaitOptions, 'timeout'>) {
```
mpeyper, line 13: could this just be a constructor on `TimeoutError`?

src/asyncUtils.ts
```diff
@@ -1,28 +1,35 @@
...
-  const waitForNextUpdate = async (options = {}) => {
+  const waitForNextUpdate = async (options: Pick<WaitOptions, 'timeout'> = {}) => {
     if (!nextUpdatePromise) {
       nextUpdatePromise = new Promise((resolve, reject) => {
-        let timeoutId
-        if (options.timeout > 0) {
-          timeoutId = setTimeout(
+        let timeoutId: number
+        if (options.timeout && options.timeout > 0) {
+          timeoutId = window.setTimeout(
```
mpeyper, line 32: why is `window` required here but not on line 20?

src/asyncUtils.ts
```diff
@@ -80,40 +91,22 @@ function asyncUtils(addResolver) {
     }
   }
 
-  const waitForValueToChange = async (selector, options = {}) => {
+  const waitForValueToChange = async (selector: () => unknown, options = {}) => {
```
mpeyper, line 94: I think `options` should be typed to `WaitOptions` here to have better support for consumers.

src/asyncUtils.ts
```diff
@@ -1,28 +1,35 @@
...
-  const timeoutError = new Error(`Timed out in ${utilName} after ${timeout}ms.`)
-  timeoutError.timeout = true
+export interface WaitOptions {
+  interval?: number
+  timeout?: number
+  suppressErrors?: boolean
+}
+
+class TimeoutError extends Error {
+  timeout = true
```
mpeyper, line 10: I don't think this field is being used anymore now that `error instanceof TimeoutError` works

mpeyper (commented):
`asyncUtils` is looking ok.  Still more disabled lint rules than I generally like.  I may have to be a bit lenient on them though given the codebase was never really designed with TS in mind.  We can always work on them piecemeal after these changes are merged.

One more thing: I noticed we have a reference to the `DefinitelyTyped` types in our [contributing guide](https://github.com/testing-library/react-hooks-testing-library/blob/master/CONTRIBUTING.md#update-typings).  That probably needs to reworded in some way or just removed completely as well as it won't be relevant once this is merged.

src/pure.tsx
```diff
@@ -4,15 +4,25 @@ import asyncUtils from './asyncUtils'
...
-type Props<T = any, R = any> = {
-  callback: (props: T) => R
-  hookProps: unknown
-  onError: CallableFunction
-  children: CallableFunction
+type TestHookProps<TProps, TResult> = {
+  callback: (props: TProps) => TResult
+  hookProps: TProps | undefined
+  onError: (error: Error) => void
+  children: (value: TResult) => void
```
JacobMGEvans, line 10: Awesome follow-up!

marcosvega91:
@mpeyper sorry I have closed it with a wrong force push. I ask @juhanakristian to open it again

PR #520, opened 2020-12-14
Convert to TypeScript

juhanakristian:
**What**:

Converting library to TypeScript, issue #498
Removes deprecated wait util (as per discussion in KCD Discord)
https://discord.com/channels/715220730605731931/785649901782433852/786338242101379142

Creating this PR on behalf of @tigerabrodi we and other people from KCD Discord are going to work on it as a group

**Checklist**:

- [x] Documentation updated
- [x] Tests
- [x] Ready to be merged
      

juhanakristian:
@mpeyper The issues you listed in the previous PR have been resolved and if there isn't anything else this should be ready for merging. Let us know what you think :)

mpeyper:
Thanks, I'll take a look today.

src/pure.tsx
```diff
@@ -0,0 +1,130 @@
...
+  } catch (err) {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+    if (err.then) {
+      throw err
+    } else {
+      onError(err)
+    }
+  }
+  return null
+}
```
mpeyper, line 35: The catch block was bothering me so I tried something... I _think_ it's better (it's definitely more explicit), but would like some feedback on it:

```suggestion
 
function isPromise<T>(value: unknown) : value is PromiseLike<T> {
  return typeof (value as PromiseLike<T>).then === 'function'
}

function TestHook<TProps, TResult>({
  callback,
  hookProps,
  onError,
  children
}: TestHookProps<TProps, TResult>) {
  try {
    if (hookProps) {
      children(callback(hookProps))
    } else {
      // coerce into undefined, so it maintains the previous behaviour
      children(callback((undefined as unknown) as TProps))
    }
  } catch (err: unknown) {
    if (isPromise(err)) {
      throw err as unknown
    } else {
      onError(err as Error)
    }
  }
  return null
}
```

(I'm not sure if you can do suggestions for multiple seperate blocks, so just focus on the `catch` block and `isPromise` function).

mpeyper (commented):
This is looking very close.  A couple of comments here, mostly minor stuff.

src/pure.tsx
```diff
@@ -0,0 +1,130 @@
...
+  onError,
+  children
+}: TestHookProps<TProps, TResult>) {
+  try {
+    if (hookProps) {
+      children(callback(hookProps))
+    } else {
+      // coerce into undefined, so it maintains the previous behaviour
+      children(callback((undefined as unknown) as TProps))
+    }
```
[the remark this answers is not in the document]
nobrayner: @mpeyper 🤦‍♂️ Thanks... I guess I got stuck down a rabit hole for that 😛

src/pure.tsx
nobrayner: Why is the promise error being cast back to unknown, after you went to the effort of Type Guarding it? You may as well just not Type Guard it and just keep err as the unknown type.

mpeyper: Yeah, it was _conceptually_ more correct than being more useful.  Perhaps leaving the return type as just `boolean` on `isPromise` is a bit easier.

mpeyper:
ok, it's been long enough... I'm going to release this for real now

PR #526, opened 2021-01-07
fix: ReactHooksRenderer renderHook was not expecting arguments

joshuaellis:
**What**:

ReactHooksRenderer was not expecting arguments due to `src/pure` missing those args in the `renderhook` function

**Why**:

fixes TS errors

**How**:

**Checklist**:

- [x] Documentation updated
- [x] Tests
- [x] Ready to be merged
      
### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -1,3 +1,5 @@
+import { RenderHookOptions } from './internal'
+
 export type Renderer<TProps> = {
   render: (props?: TProps) => void
   rerender: (props?: TProps) => void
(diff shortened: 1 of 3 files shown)


joshuaellis:
I tried this in my own project and got lots or warnings about this.

src/types/index.ts
```diff
@@ -1,3 +1,5 @@
+import { RenderHookOptions } from './internal'
```
mpeyper, line 1: I'd probably move this out of `internal` now

src/types/index.ts
```diff
@@ -48,7 +50,10 @@ export type RenderHook<
   AsyncUtils
 
 export interface ReactHooksRenderer {
-  renderHook: <TProps, TResult>() => RenderHook<TProps, TResult>
+  renderHook: <TProps, TResult, TOptions>(
+    callback: (props: TProps) => TResult,
+    options?: RenderHookOptions<TProps, TOptions>
+  ) => RenderHook<TProps, TResult>
```
mpeyper, line 56: yep, I can see why you might be getting warnings about this :P

mpeyper (approved):

src/types/index.ts
joshuaellis: I did wonder that, wanted to get your confirmation. i'll move it then merge it

PR #527, opened 2021-01-08
fix: Type for options in autodetected renderer was incorrectly inferring any passed type

mpeyper:
Changed `renderHook` in `ReactHooksRenderer` type to not use a generic for options but rather the react specific `RendererOptions`.

Moved `ReactHooksRenderer` into react specific types as all auto-detected renders are react renderers.  All modules are currently exporting both the shared types and the react types so this a non-breaking change.

**Checklist**:

- [x] Ready to be merged

### DIFF
--- a/src/pure.ts
+++ b/src/pure.ts
@@ -1,4 +1,4 @@
-import { ReactHooksRenderer } from './types'
+import { ReactHooksRenderer } from './types/react'
 
 const renderers = [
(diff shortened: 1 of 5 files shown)


joshuaellis:
Makes more sense, I think I was a bit unfamiliar with the TOptions generic and what it did. 👌🏼

PR #528, opened 2021-01-08
Refactor beta branch for a bit more consistency

mpeyper:
**What**:

More consistency across the codebase.

**Why**:

It started with me wanting to have a named function as the result of `createRenderHook` rather than the anonymous arrow function it is returning now to improve stack traces when debugging, but the more I looked the more I noticed we had a real mix of regular functions and arrow functions through the codebase, even within the same file.  This led me down a rabbit hole of a few other things that weren't quite right, names default exports and a few variable names.

**How**:

For functions I followed the rules:

1. regular function at the top level
2. arrow function for inner functions
3. arrow functions for anonymous functions

We only had a single `default export` left in the codebase.  There has been a growing movement in the community away from them and to just use named exports instead.  It's become my personal preference for all exports over the last few months.

Finally, `testHookProps` was previously renamed to `rendererProps` in all but one of the renderers (`native`), so I have changed that on now as well.  Also, I renamed `testHook` to `testHarness` in all renderers as it is the result of the `createTestHarness` helper.

**Checklist**:

- [x] Ready to be merged
      

There is also a lot of inconsistency in the types with respect to `interface` vs `type`, but I'll do another PR for that as I want change something else and #527 moves one of the types I need to change so I'll wait for that one to be merged also.

### DIFF
--- a/src/core/index.ts
+++ b/src/core/index.ts
@@ -1,7 +1,7 @@
 import { CreateRenderer, Renderer, RenderResult, RenderHook, RenderHookOptions } from '../types'
 import { ResultContainer } from '../types/internal'
 
-import asyncUtils from './asyncUtils'
+import { asyncUtils } from './asyncUtils'
 import { cleanup, addCleanup, removeCleanup } from './cleanup'
 
 function resultContainer<TValue>(): ResultContainer<TValue> {
--- a/src/core/index.ts
+++ b/src/core/index.ts
@@ -40,38 +40,40 @@ function resultContainer<TValue>(): ResultContainer<TValue> {
   }
 }
 
-const createRenderHook = <TProps, TResult, TOptions extends {}, TRenderer extends Renderer<TProps>>(
+function createRenderHook<TProps, TResult, TOptions extends {}, TRenderer extends Renderer<TProps>>(
   createRenderer: CreateRenderer<TProps, TResult, TOptions, TRenderer>
-) => (
-  callback: (props: TProps) => TResult,
-  options: RenderHookOptions<TProps, TOptions> = {} as RenderHookOptions<TProps, TOptions>
-): RenderHook<TProps, TResult, TRenderer> => {
-  const { result, setValue, setError, addResolver } = resultContainer<TResult>()
-  const renderProps = { callback, setValue, setError }
-  let hookProps = options.initialProps
+) {
+  return function renderHook(
+    callback: (props: TProps) => TResult,
+    options: RenderHookOptions<TProps, TOptions> = {} as RenderHookOptions<TProps, TOptions>
+  ): RenderHook<TProps, TResult, TRenderer> {
+    const { result, setValue, setError, addResolver } = resultContainer<TResult>()
+    const renderProps = { callback, setValue, setError }
+    let hookProps = options.initialProps
 
-  const { render, rerender, unmount, act, ...renderUtils } = createRenderer(renderProps, options)
+    const { render, rerender, unmount, act, ...renderUtils } = createRenderer(renderProps, options)
 
-  render(hookProps)
+    render(hookProps)
 
-  function rerenderHook(newProps = hookProps) {
-    hookProps = newProps
-    rerender(hookProps)
-  }
+    const rerenderHook = (newProps = hookProps) => {
+      hookProps = newProps
+      rerender(hookProps)
+    }
 
-  function unmountHook() {
-    removeCleanup(unmountHook)
-    unmount()
-  }
+    const unmountHook = () => {
+      removeCleanup(unmountHook)
+      unmount()
+    }
 
-  addCleanup(unmountHook)
+    addCleanup(unmountHook)
 
-  return {
-    result,
-    rerender: rerenderHook,
-    unmount: unmountHook,
-    ...asyncUtils(act, addResolver),
-    ...renderUtils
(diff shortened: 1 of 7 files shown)


src/core/index.ts
```diff
@@ -40,39 +40,48 @@ function resultContainer<TValue>(): ResultContainer<TValue> {
...
+      unmount: unmountHook,
+      ...asyncUtils(act, addResolver),
+      ...renderUtils
+    }
   }
+
+  // If the function name does not get used before it is returned,
+  // it seems to vanish in nodejs and does not appear in stack traces.
+  // This dummy usage works around that.
+  const _name = renderHook.name // eslint-disable-line @typescript-eslint/no-unused-vars
```
mpeyper, line 82: This is gross and I have no idea why it is necessary, but in node, if I've got function that returns a function like so:
```js
function createFunc() {
  function func() {
  }
  return func
}
```

then:

```js
console.log(createFunc.name) // 'createFunc'
console.log(createFunc().name) // ''
```

But if I use `func.name` inside `createFunc` like so:

```js
function createFunc() {
  function func() {
  }
  const name = func.name // not returned, just used
  return func
}
```

then:

```js
console.log(createFunc.name) // 'createFunc'
console.log(createFunc().name) // 'func'
```

In the chrome console, both example will return `func` for the second log.  It's very weird.

src/helpers/createTestHarness.tsx
```diff
@@ -39,4 +39,13 @@ export const createTestHarness = <TProps, TResult>(
     }
     return component
   }
+
+  // If the function name does not get used before it is returned,
+  // it seems to vanish in nodejs and does not appear in stack traces.
+  // This dummy usage works around that.
+  const _name = testHarness.name // eslint-disable-line @typescript-eslint/no-unused-vars
```
mpeyper, line 46: [same excuse](https://github.com/testing-library/react-hooks-testing-library/pull/528/files#r553757160)

src/core/index.ts
joshuaellis: That is really weird. Can't seem to find much online either how to potentially solve.

mpeyper: So I've discovered that this isn't a node thing, but something in the compilation.  I've narrowed it down to babel (not TS and not something in `jest`) as if I use `kcd-scripts` to compile a javascript only file, this

```js
export function createFunc1() {
  function func() {
  }
  return func
}

export function createFunc2() {
  function func() {
  }
  const name = func.name // not returned, just used
  return func
}
```

becomes 

```js
"use strict";

Object.defineProperty(exports, "__esModule", {
  value: true
});
exports.createFunc1 = createFunc1;
exports.createFunc2 = createFunc2;

function createFunc1() {
  return function () {};
}

function createFunc2() {
  function func() {}

  func.name; // not returned, just used

  return func;
}
```

The name in `createFunc1` has been scrubbed!

Now I just have to work out which plugin is causing this.

jpeyper: its babel-plugin-minify-dead-code-elimination

https://babeljs.io/en/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=usage&spec=false&loose=false&code_lz=GYVwdgxgLglg9mABBATgUwIZTQMXBACgEpEBvAKEUVElgWv2LMsQF8X0oQUkaJz25CAgDOcADZoAdOLgBzAqkzY8kKWAwBbNESGiJ02QqVZcjIuq06gA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=true&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Ces2015%2Creact%2Ctypescript&prettier=true&targets=&version=7.12.12&externalPlugins=babel-plugin-minify-dead-code-elimination%400.5.1

jpeyper, line 82: you dont need to assign the name to a variable to have have it not be dead code eliminated.

```suggestion
  renderHook.name
```

jpeyper, line 82: you dont need to assign the name to a variable to have have it not be dead code eliminated.

```suggestion
  // If the function name does not get used before it is returned,
  // it's name is removed by babel-plugin-minify-dead-code-elimination
  // This dummy usage works around that.
  renderHook.name
```

src/helpers/createTestHarness.tsx
jpeyper, line 46: [same suggestion](https://github.com/testing-library/react-hooks-testing-library/pull/528/files#r553863248)

```suggestion
  // If the function name does not get used before it is returned,
  // it's name is removed by babel-plugin-minify-dead-code-elimination
  // This dummy usage works around that.
  testHarness.name
```

src/core/index.ts
mpeyper: Thanks @jpeyper.  I've raised [an issue in `kcd-scripts`](https://github.com/kentcdodds/kcd-scripts/issues/188) about this.

mpeyper: I'll keep an eye on the other issue, in the meantime, we have this workaround 🤷

joshuaellis (approved):
Looks good, interested to hear the feedback on kcd-scripts RE minify plugin.

mpeyper:
@allcontributors please add @jpeyper for review, code

PR #529, opened 2021-01-08
Release/5.0.0 docs

joshuaellis:
**What**:

* Added a new section to the docs regarding the renderer options now availabe.
* Updated introduction and readme (as they're in sync)
* Updated API reference to include `hydrate` and re-word `act` to reflect renderer
* Update snippets in API reference

**Why**:

Compliments the changes merged in #510 for #68 

**How**:

`Docz`

**Checklist**:

- [x] Documentation updated
- [x] Ready to be merged


doczrc.js
```diff
@@ -19,6 +19,7 @@ export default {
   menu: [
     { name: 'Introduction' },
     { name: 'Usage', menu: ['Basic Hooks', 'Advanced Hooks'] },
+    { name: 'Renderer' },
```
joshuaellis, line 22: I wasn't sure where to put this, but I felt this probably was a good place...

docs/renderer.md
```diff
@@ -0,0 +1,61 @@
...
+import { renderHook, act } from '@testing-library/react-hooks' // will try to auto-detect
+
+import { renderHook, act } from '@testing-library/react-hooks/dom' // will use react-dom
+
+import { renderHook, act } from '@testing-library/react-hooks/native' // will use react-test-renderer
+
+import { renderHook, act } from '@testing-library/react-hooks/server' // will use react-dom
+```
+
+## Caveats
```
joshuaellis, line 48: I don't believe i've missed anything that's worth going here?

docs/introduction.md
```diff
@@ -68,17 +68,21 @@ npm install --save-dev @testing-library/react-hooks
 ### Peer Dependencies
 
 `react-hooks-testing-library` does not come bundled with a version of
-[`react`](https://www.npmjs.com/package/react) or
-[`react-test-renderer`](https://www.npmjs.com/package/react-test-renderer) to allow you to install
-the specific version you want to test against. Generally, the installed versions for `react` and
-`react-test-renderer` should have matching versions:
+[`react`](https://www.npmjs.com/package/react) to allow you to install the specific version you want
```
joshuaellis, line 71: I wonder if this could be made more concise, I think at least for now it gets the point across though.

doczrc.js
mpeyper: Part of me wonders if this and the [Installation instructions](https://react-hooks-testing-library.com/#installation) section should be combined into an `Installation` page?

docs/api-reference.md
```diff
@@ -11,16 +11,17 @@ route: '/reference/api'
 - [`act`](/reference/api#act)
 - [`cleanup`](/reference/api#cleanup)
 - [`addCleanup`](/reference/api#addcleanup)
+- [`removeCleanup`](/reference/api#removecleanup)
```
mpeyper, line 14: Whoops

docs/api-reference.md
```diff
@@ -11,16 +11,17 @@ route: '/reference/api'
...
 
-```js
-function renderHook(
-  callback: function(props?: any): any,
+```ts
+renderHook: (
+  callback: (props?: any): any,
   options?: RenderHookOptions
-): RenderHookResult
+) => RenderHookResult
```
mpeyper, line 24: I'm not sure these read better or easier than the old ones, especially for non-TS users.

docs/api-reference.md
```diff
@@ -77,22 +78,33 @@ returned at the time.
...
-```js
-function unmount(): void
+```ts
+unmount: () => void
 ```
 
 A function to unmount the test component. This is commonly used to trigger cleanup effects for
 `useEffect` hooks.
 
+### `hydrate`
```
mpeyper, line 97: I know you've got the message about this only being available in the server renderer, but I wonder if a seperate section for SSR would make it even more apparent?

docs/api-reference.md
```diff
@@ -77,22 +78,33 @@ returned at the time.
...
+
+```ts
+hydrate: () => void
+```
+
+> This is only used when using the `server` module. See [Rendering](/renderer) for more information
+
+A function to hydrate the component. This is commonly used before `act` calls and after `rerender`
```
mpeyper, line 105: I think it would have to be before ` rerender` too.

docs/renderer.md
```diff
@@ -0,0 +1,61 @@
+---
+name: Renderer
+route: '/renderer'
+---
+
+# Render Engine
+
+## Overview
+
+React requires a rendering engine, typically when creating an application people use `react-dom`.
```
mpeyper, line 10: Perhaps a mention of `react-native` using `react-test-renderer` as the defacto renderer for tests (at least I think it is)?

docs/renderer.md
```diff
@@ -0,0 +1,61 @@
...
+import { renderHook } from '@testing-library/react-hooks'
+```
+
+It does this because the library runs a check for available renderers, in the order:
+
+- `react-test-renderer`
+- `react-dom`
+
+If neither are available you will see an error asking you to check that one of the above is
+installed. If only one is installed, then it will use that renderer.
```
mpeyper, line 25: The part here about how we auto detect and errors feels a bit too specific to our implementation details.  It probable enough to say we will auto detect a default renderer if either of the dependencies is installed and will use, `react-test-renderer` if both are installed.

Then maybe a note about why that's the default, i.e. fewer dependencies on the test environment (i.e. no DOM requirement) and works for hooks intended for both `react` and `react-native`.

docs/renderer.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Being specific
+
+If, however, for certain tests you want to use a specific renderer (e.g. you want to use `react-dom`
+for SSR) you can import the server module directly:
```
mpeyper, line 30: I like the "Being specific" section, but I think we could have a section (or page?) on SSR where we can talk specifically about hydration and any other server specific quirks.

Thoughts?

docs/renderer.md
mpeyper: The only things I can think of are:

1. Auto detect may not work if tests get bundle to run in the browser.
2. State won't update with the `server` renderer until after `hydrate` is called.

docs/renderer.md
mpeyper: Doh, it's there, just lower in the file. Ignore this (as although maybe switch the order?)

docs/renderer.md
```diff
@@ -0,0 +1,61 @@
...
+### SSR
+
+While calling `renderHook` from `@testing-library/react-hooks/native` and
+`@testing-library/react-hooks/dom` will return the same `RenderHookResult` as documented
+[here](/reference/api#renderhook-result), using `@testing-library/react-hooks/server` will return an
+additional function:
+
+```ts
+hydrate: () => void
+```
```
mpeyper, line 59: As mentioned above, perhaps some fluff about when `hydrate` is necessary

docs/api-reference.md
joshuaellis: > non-TS users

I think that's fair, I didn't consider that.

doczrc.js
joshuaellis: That sounds like a good idea, i've probably repeated myself across in `Peer Dependencies` & `Renderer` anyway.

docs/renderer.md
```diff
@@ -0,0 +1,66 @@
...
+import { renderHook, act } from '@testing-library/react-hooks/dom' // will use react-dom
+
+import { renderHook, act } from '@testing-library/react-hooks/native' // will use react-test-renderer
+
+import { renderHook, act } from '@testing-library/react-hooks/server' // will use react-dom
+```
+
+## Caveats
+
+## Auto detect
```
mpeyper, line 50: should this be `###`?

docs/renderer.md
joshuaellis: I think SSR you're right, own page. thinking it should go under usage, underneath advanced hooks.

README.md
```diff
@@ -141,17 +142,21 @@ npm install --save-dev @testing-library/react-hooks
...
-[`react-test-renderer`](https://www.npmjs.com/package/react-test-renderer) to allow you to install
-the specific version you want to test against. Generally, the installed versions for `react` and
-`react-test-renderer` should have matching versions:
+[`react`](https://www.npmjs.com/package/react) to allow you to install the specific version you want
+to test against. It also does not come installed with a specific renderer, we currently support
+[`react-test-renderer`](https://www.npmjs.com/package/react-test-renderer) and
+[`react-dom`](https://www.npmjs.com/package/react-dom), you only need to install one of them.
+However,you can install both, but `react-test-renderer` is the default, for more information see the
+[installation docs](https://react-hooks-testing-library.com/#installation). Generally, the installed
+versions for `react` and the selected renderer should have matching versions:
```
mpeyper, line 151: Missing a space after the comma and a few typos here, but I've got a suggestion for updating the grammar a bit anyway:

```suggestion
[`react-dom`](https://www.npmjs.com/package/react-dom). You only need to install one of them,
however, if you do have both installed, we will use `react-test-renderer` as the default. For more
information see the [installation docs](https://react-hooks-testing-library.com/#installation).
Generally, the installed versions for `react` and the selected renderer should have matching
versions:
```

docs/installation.md
```diff
@@ -0,0 +1,81 @@
...
+If you have both installed in your project, assuming you're using the default imports (see below)
+the library defaults to using `react-test-renderer`. This is because rtr is the defacto test
+renderer for `react-native` for tests and it enables hooks to be correctly tested that are written
+for both `react` & `react-native`.
+
+```js
+import { renderHook } from '@testing-library/react-hooks'
+```
+
+> The auto detection function may not work if tests are bundled to run in the browser.
```
mpeyper, line 54:
````suggestion
When running tests, a renderer is required in order to render the React component we wrap around
your hook. We currently support two different renderers:

- `react-test-renderer`
- `react-dom`

When using standard import for this library (show below), we will attempt to auto-detect which
renderer you have installed and use it without needing any specific wiring up to make it happen. If
you have both installed in your project, and you use the standard import (see below) the library
will default to using `react-test-renderer`.

> We use `react-test-renderer` by default as it enables hooks to be tested that are designed for
> either `react` or `react-native` and it is compatible with more test runners out-of-the-box as
> there is no DOM requirement to use it.

The standard import looks like:

```js
```

> Note: The auto detection function may not work if tests are bundles before execution (e.g. to be
> run in a browser)

### Act

Each render also provides a unique [`act` function](https://reactjs.org/docs/test-utils.html#act)
that cannot be used with other renderers. In order to simplify with `act `function you need to use,
we also export the correct one alongside the detected renderer for you:

```js
import { renderHook, act } from '@testing-library/react-hooks'
```
````

File: docs/installation.md [hunk 15]
```diff
@@ -0,0 +1,81 @@
...
+
+```ts
+import { renderHook, act } from '@testing-library/react-hooks' // will try to auto-detect
+
+import { renderHook, act } from '@testing-library/react-hooks/dom' // will use react-dom
+
+import { renderHook, act } from '@testing-library/react-hooks/native' // will use react-test-renderer
+
+import { renderHook, act } from '@testing-library/react-hooks/server' // will use react-dom
+```
```
mpeyper, line 75:
````suggestion
Auto-detection is great for simplifying setup and getting out of your way, but sometimes you do need
a little but more control. If a test needs requires a specific type of environment, the import can
be appended to force a specific renderer to be use. The supported environments are:

- `dom`
- `native`
- `server`

The imports for each type of renderer are as follows:

```ts
import { renderHook, act } from '@testing-library/react-hooks' // will attempt to auto-detect

import { renderHook, act } from '@testing-library/react-hooks/dom' // will use react-dom

import { renderHook, act } from '@testing-library/react-hooks/native' // will use react-test-renderer

import { renderHook, act } from '@testing-library/react-hooks/server' // will use react-dom/server
```
````

**Comment by joshuaellis** [comment]
Those are all the suggestion added, thanks for the grammer sweep, not sure if you wanna have one more glance over on the newest commit?

**Comment by mpeyper** [comment]
Will do, but in the morning (passed midnight here now... I've been "finishing up for the night" for about 2 hours now 😅)

**Comment by joshuaellis** [comment]
>  I've been "finishing up for the night" for about 2 hours now

I know that feeling! no rush of course lol.

File: docs/api-reference.md [hunk 16]
```diff
@@ -86,13 +87,26 @@ passed, they will replace the `callback` function's `initialProps` for subsequen
...
+```ts
+function hydrate(): void
+```
+
+> This is only used when using the `server` module. See [SSR](/usage/ssr-hooks) for more information
+> on server-side rendering your hooks.
+
+A function to hydrate the component. This is required before you can interact with the hook, whether
+that is an `act` or `rerender` call. Some effects such as `useEffect` will not be called unless
+`hydrate` is called.
```
mpeyper, line 108:
```suggestion
A function to hydrate a server rendered component into the DOM. This is required before you can
interact with the hook, whether that is an `act` or `rerender` call. Effects created using
`useEffect` or `useLayoutEffect` are also not run on server rendered hooks until `hydrate` is
called.
```

docs/api-reference.md
```diff
@@ -102,15 +116,15 @@ Utilities to assist with testing asynchronous behaviour. See the
 
 ## `act`
 
-This is the same [`act` function](https://reactjs.org/docs/test-utils.html#act) that is exported by
-`react-test-renderer`.
+This is the same [`act` function](https://reactjs.org/docs/test-utils.html#act) function that is
+exported from your chosen renderer. Although, they both work the same.
```
mpeyper, line 120: That last sentence reads a bit odd.  I'd just drop it all together.

```suggestion
exported from your chosen renderer.
```

docs/api-reference.md
```diff
@@ -142,7 +156,8 @@ module.exports = {
 ```
 
 Alternatively, you can change your test to import from `@testing-library/react-hooks/pure` instead
-of the regular imports.
+of the regular imports. This applys to any of our export methods documented in
+[Rendering](/rendering#being-specific).
```
mpeyper, line 160: dead link

```suggestion
[Rendering](/installation#being-specific).
```

docs/api-reference.md
```diff
@@ -11,14 +11,15 @@ route: '/reference/api'
...
 
 ---
 
 ## `renderHook`
 
-```js
+```ts
 function renderHook(
-  callback: function(props?: any): any,
+  callback: (props?: any): any,
```
mpeyper, line 22: I don't think this syntax is valid.  This matches what you've got below for other callbacks

```suggestion
  callback: (props?: any) => any,
```

docs/usage/ssr-hooks.md
```diff
@@ -0,0 +1,96 @@
...
+function hydrate(): void
+```
+
+The `hydrate` function is a light wrapper around
+[`ReactDOM.hydrate`](https://reactjs.org/docs/react-dom.html#hydrate) but no arguments are required
+as the library will pass the element & container for you. Remember, certain effects such as
+`useEffect` will not run server side and `hydrate` must be called before those effects are
+ran.`hydrate`is also necessary before the first `act` or `rerender` call. For more information on
+`hydrate` see the [API documentation](/reference/api#hydrate). There is also an
+[example below](/usage/ssr-hooks#example)
```
mpeyper, line 33:
````suggestion
---
name: Server-Side Rendering
menu: Usage
route: '/usage/ssr-hooks'
---

# Server-Side Rendering (SSR)

## Setup

To test how your hook will behave when rendered on the server, you can change your import to the use
the `server` module:

```ts
import { renderHook } from '@testing-library/react-hooks/server'
```

> SSR is only available when using the `react-dom` renderer. Please refer to the
> [installation guide](/installation#peer-dependencies) for instructions and supported versions.

This import has the same [API as the standard import](/reference/api) except the behaviour changes
to use SSR semantics.
````

File: docs/usage/ssr-hooks.md [hunk 21]
```diff
@@ -0,0 +1,96 @@
...
+
+    // hydrate is called because we want to interact with the hook
+    hydrate()
+
+    act(() => result.current.decrement())
+
+    expect(result.current.count).toBe(-1)
+  })
+})
+```
```
mpeyper, line 65:
````suggestion
## Hydration

The result of rendering you hook is static are not interactive until it is hydrated into the DOM.
This can be done using the `hydrate` function that is returned from `renderHook`.

Consider the `useCounter` example from the [Basic Hooks section](/usage/basic-hooks):

```js
import { useState, useCallback } from 'react'

export default function useCounter() {
  const [count, setCount] = useState(0)
  const increment = useCallback(() => setCount((x) => x + 1), [])
  return { count, increment }
}
```

If we try to call `increment` immediately after server rendering, nothing happens and the hook is
not interactive:

```js
import { renderHook, act } from '@testing-library/react-hooks/server'
import useCounter from './useCounter'

test('should increment counter', () => {
  const { result } = renderHook(() => useCounter(0))

  act(() => {
    result.current.increment()
  })

  expect(result.current.count).toBe(1) // fails as result.current.count is still 0
})
```

We can make the hook interactive by calling the `hydrate` function that is returned from
`renderHook`:

```js
import { renderHook, act } from '@testing-library/react-hooks/server'
import useCounter from './useCounter'

test('should increment counter', () => {
  const { result, hydrate } = renderHook(() => useCounter(0))

  hydrate()

  act(() => {
    result.current.increment()
  })

  expect(result.current.count).toBe(1) // now it passes
})
```

Anything that causes the hook's state to change will not work until `hydrate` is called. This
includes both the [`rerender`](http://localhost:3000/reference/api#rerender) and
[`unmount`](http://localhost:3000/reference/api#unmount) functionality.
````

File: docs/usage/ssr-hooks.md [hunk 22]
```diff
@@ -0,0 +1,96 @@
...
+    expect(sideEffect[1]).toBe(false)
+    expect(sideEffect[2]).toBe(false)
+
+    hydrate()
+
+    expect(sideEffect[1]).toBe(true)
+    expect(sideEffect[2]).toBe(false)
+  })
+})
+```
```
mpeyper, line 96:
````suggestion
Another caveat of SSR is that `useEffect` and `useLayoutEffect` hooks, by design, do not run on when
rendering.

Consider this `useTimer` hook:

```js
import { useState, useCallback, useEffect } from 'react'

export default function useTimer() {
  const [count, setCount] = useState(0)
  const reset = useCallback(() => setCount(0), [])
  useEffect(() => {
    const intervalId = setInterval(() => setCount((c) => c + 1, 1000))
    return () => {
      clearInterval(intervalId)
    }
  })
  return { count, reset }
}
```

Upon initial render, the interval will not start:

```js
import { renderHook, act } from '@testing-library/react-hooks/server'
import useTimer from './useTimer'

test('should start the timer', async () => {
  const { result, waitForValueToChange } = renderHook(() => useTimer(0))

  await waitForValueToChange(() => result.current.count) // times out as the value never changes

  expect(result.current.count).toBe(1) // fails as result.current.count is still 0
})
```

Similarly to updating the hooks state, the effect will start after `hydrate` is called:

```js
import { renderHook, act } from '@testing-library/react-hooks/server'
import useTimer from './useTimer'

test('should start the timer', async () => {
  const { result, hydrate, waitForValueToChange } = renderHook(() => useTimer(0))

  hydrate()

  await waitForValueToChange(() => result.current.count) // now resolves when the interval fires

  expect(result.current.count).toBe(1)
})
```
````

File: docs/usage/ssr-hooks.md [hunk 23]
```diff
@@ -0,0 +1,96 @@
+---
+name: SSR Hooks
+menu: Usage
+route: '/usage/ssr-hooks'
```
mpeyper, line 4: I think the route and the file should become just `/usage/ssr` (I haven't done it as a suggestion because I can't rename the file to match).

mpeyper:
I think that last round of suggestions will just about do it.  Good enough for the first release anyway.  If history is any indicator, there will be plenty of PRs suggesting minor edits to them anyway, regardless of how much effort we put in now 😅

docs/usage/ssr-hooks.md
mpeyper: If you do this, don't forger to update any links to this page (I saw at least one in `api-reference`)

PR #531, opened 2021-01-09
refactor: move tests to specificly related folders

joshuaellis:
**What**:

Proposal to move test files to be closer to the code that we're testing, and rename those test files to be `*.test.ts(x)`.

**Why**:

This was discussed in #510, I'd say the main issue is they're all named the same and incredibly close together, now we're maintaining "three engines" we may mind that the testing needs to diverge, an example is seen in `src/server` where we need to test the hydration warnings. It also, imo, makes the repo look tidier.

Renaming the tests was a personal choice, if i'm jumping around a lot of files, it's really helpful to understand whats a test file just from the name. Added bonus, you normally get a different icon in your IDE.

Also we do want to try and figure out how to cover `src/pure.ts` so this wouldn't actually fit into the old test folder pattern because they were related to the renderer. Therefore when we cross that bridge I would suggest putting the test in `src/__tests__` and thus naming it `pure.test.ts`.

**Checklist**:

- [x] Tests
- [x] Ready to be merged


joshuaellis:
hm. probably need to move that `tsconfig.json` out of `src/__tests__` and just into `test`.... that's annoying.

src/dom/__tests__/asyncHook.test.ts
```diff
@@ -1,5 +1,5 @@
 import { useState, useRef, useEffect } from 'react'
-import { renderHook } from '../../src/dom'
+import { renderHook } from '../index'
```
mpeyper, line 2: I think `import { renderHook } from '..'` works too and might be a bit easier to maintain

src/__tests__/tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
-  "extends": "../tsconfig",
+  "extends": "../../tsconfig",
   "compilerOptions": {
     "declaration": false
   },
   "exclude": [],
-  "include": ["./**/*.ts"]
+  "include": ["**/*.test.(ts|tsx)"]
```
mpeyper, line 7: Removing the `.test` from the files names was explicit in the `kcd-scripts` migration (#502), but I'm no bothered and happy to keep them if you prefer it.

src/__tests__/tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
```
mpeyper, line 1: I don't think this file is actually needed anymore if you remove the `exclude` from the root `tsconfig.json`

jest.config.js
```diff
@@ -2,6 +2,6 @@
 const { jest: jestConfig } = require('kcd-scripts/config')
 
 module.exports = Object.assign(jestConfig, {
-  roots: ['<rootDir>/src', '<rootDir>/test'],
-  testMatch: ['<rootDir>/test/**/*.(ts|tsx|js)']
+  roots: ['<rootDir>/src'],
+  testMatch: ['<rootDir>/**/__tests__/*.(ts|tsx|js)']
```
mpeyper, line 6: I think this file can just be deleted now as these match the default jest config in `kcd-scripts`

tsconfig.json
```diff
@@ -3,5 +3,6 @@
   "compilerOptions": {
     "allowJs": true,
```
mpeyper, line 4: Technically, I think this setting can be removed now.  It was only there while the codebase was being converted, but now everything should be TS going forward.

mpeyper (approved):
A few suggestions, but no showstoppers

mpeyper:
You have forgotten to updated [the eslintrc](https://github.com/testing-library/react-hooks-testing-library/blob/beta/.eslintrc#L14) file which is why the CI is failing.

mpeyper:
> hm. probably need to move that `tsconfig.json` out of `src/__tests__` and just into `test`.... that's annoying.

If my other comments are right, you can just delete it

src/dom/__tests__/asyncHook.test.ts
joshuaellis: sure thing 👍🏼

joshuaellis:
Thanks for doing those updates.

PR #537, opened 2021-01-12
Refactor async utils

mpeyper:
**What**:

Completely refactor the async utils to be more correct and have more sensible defaults.

**Why**:

This is something I've wanted to do for a while now.  `timeout` and `interval` were both added after the async utils were originally developed, so fo backwards compatibility, we left them as defaulting to off. The catalyst for this change was to add in defaults and require people to opt out of the functionality if they wanted to.

I thought it was going to be straight forward to add the default, but when I did I found lots (most) of our async tests began failing.  I discovered that we actually had a few bugs in the implementation, particularly around overlapping `act` calls and seemingly a leaking promise that wasn't being `awaited` in some circumstances.  So I decided to build it from scratch and see how that went.

**How**:

The main change of the refactor is restructuring thee order of the utils.  Previously, `waitForNextUpdate` was the base utility that all others were build off.  `waitFor` piggy backed off of the render updates and added a layer of interval checking over the top. `waitForValueToChange` was a thing wrapper around `waitFor` to turn a selector into a callback.  In the new version, there is a base `wait` util that is not exported, but provides the render updates, interval checking and timeouts, but with minimal logic around what to do other than call the callback.  All of the other utilities now provide a thin wrapper around `wait` to add their pieces of nuanced logic. 

Because the default behaviour of `waitFor` was to suppress any errors (so `expect` could be used in the callback), there was a really clunky option called `suppressErrors` that `waitForValueToChange` would override to throw the errors if a bad selector was used.  It was annoying to document, confusing when to set it and I don't believe anyone should ever want to use anything but the defaults.  A consequence of this refactor was I was able to remove `suppressErrors` without losing the functionality.

The other change is that you can no longer `await` multiple utilities at the same time without getting a warning about having overlapping `act` calls.  I'm not sure why I felt it was so important to support this use case before I can't see any reason why anyone would need to do this.  removing it simplified the internal `act` logic and  the overall complexity of the base `wait` utility.

**Checklist**:

- [x] Documentation updated
- [x] Tests
- [x] Ready to be merged

**Note: this is a breaking change**

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -201,7 +201,7 @@ removed, the provided callback will no longer execute as part of running
 ### `waitForNextUpdate`
 
 ```ts
-function waitForNextUpdate(options?: { timeout?: number }): Promise<void>
+function waitForNextUpdate(options?: { timeout?: number | false }): Promise<void>
 ```
 
 Returns a `Promise` that resolves the next time the hook renders, commonly when state is updated as
@@ -209,7 +209,9 @@ the result of an asynchronous update.
 
 #### `timeout`
 
-The maximum amount of time in milliseconds (ms) to wait. By default, no timeout is applied.
+_Default: 1000_
+
+The maximum amount of time in milliseconds (ms) to wait.
 
 ### `waitFor`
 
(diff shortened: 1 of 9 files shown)


mpeyper:
I wonder if these changes fix the `act` issues I was having when trying to implement a `preact` renderer?

joshuaellis (approved):
All looks good, that might fix some of the issues i've been having trying to help someone discord with some odd hooks (that they've written)

Issue #50, opened 2019-04-24
regression: console.error supression no longer works in 0.5


ntucker:
- `react-hooks-testing-library` version: 0.5.0
- `react-testing-library` version: 6.1.2
- `react` version: 16.8.6
- `node` version: 11.10
- `npm` (or `yarn`) version: 1.15.2

### Relevant code or config:

https://github.com/facebook/react/issues/11098#issuecomment-412682721

```js
let your = (code, tell) => `the ${story}`;
```

### What you did:

```
function onError(e: any) {
  e.preventDefault();
}
beforeEach(() => {
  window.addEventListener('error', onError);
});
afterEach(() => {
  window.removeEventListener('error', onError);
});
```

### What happened:
console.error node_modules/react-test-renderer/cjs/react-test-renderer.development.js:9215
    The above error occurred in the <TestHook> component:
        in TestHook
        in Suspense
        in ErrorBoundary
        in Suspense (created by wrapper)
        in ExternalCacheProvider
        in Unknown (created by wrapper)
        in wrapper
    
    React will try to recreate this component tree from scratch using the error boundary you provided, ErrorBoundary.

### Problem description:
This used to suppress the console errors from things thrown in render being caught by errorboundaries. It doesn't work anymore

mpeyper:
I'll take a look, but the big difference between the versions was changing from `react-testing-library` (which uses `react-dom` under the hood) to using `react-test-renderer` to handle rendering, and I'm not sure if the fix mentioned in Dan's comment applies to both renderers, or just `react-dom`.

ntucker:
Hmm, I wonder what the expected workaround is for react-test-renderer....one would think it shouldn't even trigger this since it's definitely in testing mode.

mpeyper:
I had a look at the source code for `resct-test-renderer` and the fix is there, but I think it might have a bug. 

Putting console logs into their error handler where it checks for a prevented default (which is added as an event listener to `window`, same as yours) never logs, but a log where it checks for the suppression flag does log.

This tells be that the way errors are handled in the test renderer doesn't trigger window events, which sort of makes sense as the point is it doesn't require a DOM to render.

I'm still investigating, but I see 3 options ahead:

1. We do nothing and raise a bug with react (assuming it's their bug)
2. We use `react-dom` (probably with `react-testing-library` again) for renderer information
3. We don't use an error boundary for error handling and catch the errors in our test component

mpeyper:
[...]

PR #539, opened 2021-01-13
feat: use error boundary to capture useEffect errors

### LINKED ISSUE #308: Errors handing for effect callbacks
Originally raised in #305, errors throw in `useEffect` (and `useLayoutEffect`) callbacks are not being captured in `result.error` like other rendering and async errors.

A quick test showed that using an error boundary would catch the errors in the effects. We used to have a an error boundary for error handling, but after issues found in #50 and #74, it was removed.

I'm proposing we reintroduce the error back into the mix.

As an alternative approach, I think it might also worth exploring an alternative where we no longer try to catch errors and just have `renderHook`, `rerender` and all the async utils just throw them.  This would be a breaking change, but I suspect it will actually be a more ergonomic and idiomatic testing experience.  I haven't put too much thought into this approach yet, so there might be some issues with it.

mpeyper:
**What**:

Replace `try/catch` error handling with a an error boundary.

**Why**:

Fixes #308

The error boundary will catch errors thrown in `useEffect` calls as well as in the render path.

**How**:

The challenge here and why we haven't done it sooner is the excessive amount of logging produced from React when an error boundary catches and error.  I've tried many official and unofficial approaches over the years but none have worked consistently as we changed renderers and new versions of React were released. For more information, see these issues:
  - https://github.com/testing-library/react-hooks-testing-library/issues/50
  - https://github.com/facebook/react/issues/11098#issuecomment-412682721
  - https://github.com/facebook/react/issues/15520
  - https://github.com/facebook/react/issues/18841

In this proposal, I used a package called [filter-console](https://www.npmjs.com/package/filter-console) to prevent errors matching a specific patterns from being logged.  This is potentially brittle if new versions of React change the log format, but we can add more patterns or change the existing ones if they ever do 🤷‍♂️ 

I also opted to use `react-error-boundary` as it provided some nice features for resetting the error state when rerendering.  The way I access the reset functionality is a bit clunky, so I'm open to suggestions on how to clean that up.

**Checklist**:

- [x] Tests
- [x] Ready to be merged
      

We had a bunch of tests marked as `.skip` for testing the effect errors that I have unmarked and they all work now, which is a good sign.

The other tests that needed updating was the `result history tests` .  The hook used in those tests actually had a side effect from rendering that the `try/catch` was hiding, but the error boundary showed the results appearing out of order because the [error handling triggered an additional render](https://github.com/facebook/react/issues/16130#issuecomment-521637592) that cause the next result to jump ahead in the array.  It was painful to debug, but the new tests worked against the old implementation and as well as the new one.

### DIFF
--- a/src/helpers/createTestHarness.tsx
+++ b/src/helpers/createTestHarness.tsx
@@ -1,42 +1,65 @@
 import React, { Suspense } from 'react'
+import { ErrorBoundary, FallbackProps } from 'react-error-boundary'
+import filterConsole from 'filter-console'
+
+import { addCleanup } from '../core'
 
 import { RendererProps, WrapperComponent } from '../types/react'
 
-import { isPromise } from './promises'
-
-function TestComponent<TProps, TResult>({
-  hookProps,
-  callback,
-  setError,
-  setValue
-}: RendererProps<TProps, TResult> & { hookProps?: TProps }) {
-  try {
-    // coerce undefined into TProps, so it maintains the previous behaviour
-    setValue(callback(hookProps as TProps))
-  } catch (err: unknown) {
-    if (isPromise(err)) {
-      throw err
-    } else {
-      setError(err as Error)
+function suppressErrorOutput() {
+  // The error output from error boundaries is notoriously difficult to suppress.  To save
+  // out users from having to work it out, we crudely suppress the output matching the patterns
+  // below.  For more information, see these issues:
+  //   - https://github.com/testing-library/react-hooks-testing-library/issues/50
+  //   - https://github.com/facebook/react/issues/11098#issuecomment-412682721
+  //   - https://github.com/facebook/react/issues/15520
+  //   - https://github.com/facebook/react/issues/18841
+  const removeConsoleFilter = filterConsole(
+    [
+      /^The above error occurred in the <TestComponent> component:/, // error boundary output
+      /^Error: Uncaught .+/ // jsdom output
+    ],
+    {
+      methods: ['error']
     }
-  }
-  return null
+  )
+  addCleanup(removeConsoleFilter)
 }
 
 function createTestHarness<TProps, TResult>(
-  rendererProps: RendererProps<TProps, TResult>,
+  { callback, setValue, setError }: RendererProps<TProps, TResult>,
   Wrapper?: WrapperComponent<TProps>,
   suspense: boolean = true
 ) {
+  const TestComponent = ({ hookProps }: { hookProps?: TProps }) => {
+    // coerce undefined into TProps, so it maintains the previous behaviour
+    setValue(callback(hookProps as TProps))
+    return null
+  }
+
+  let resetErrorBoundary = () => {}
+  const ErrorFallback = ({ error, resetErrorBoundary: reset }: FallbackProps) => {
+    resetErrorBoundary = () => {
+      resetErrorBoundary = () => {}
+      reset()
+    }
+    setError(error)
+    return null
+  }
+
+  suppressErrorOutput()
+
   const testHarness = (props?: TProps) => {
-    let component = <TestComponent hookProps={props} {...rendererProps} />
+    resetErrorBoundary()
+
+    let component = <TestComponent hookProps={props} />
     if (Wrapper) {
       component = <Wrapper {...(props as TProps)}>{component}</Wrapper>
     }
     if (suspense) {
       component = <Suspense fallback={null}>{component}</Suspense>
     }
-    return component
+    return <ErrorBoundary FallbackComponent={ErrorFallback}>{component}</ErrorBoundary>
   }
 
   return testHarness
--- a/src/helpers/promises.ts
+++ b/src/helpers/promises.ts
@@ -2,13 +2,9 @@ function resolveAfter(ms: number) {
   return new Promise<void>((resolve) => setTimeout(resolve, ms))
 }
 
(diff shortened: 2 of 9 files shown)


src/helpers/promises.ts
```diff
@@ -2,13 +2,9 @@ function resolveAfter(ms: number) {
   return new Promise<void>((resolve) => setTimeout(resolve, ms))
 }
 
-export async function callAfter(callback: () => void, ms: number) {
+async function callAfter(callback: () => void, ms: number) {
```
mpeyper, line 5: moved the the export at the end of the file

src/server/pure.ts
```diff
@@ -20,8 +20,12 @@ function createServerRenderer<TProps, TResult>(
...
       renderProps = props
       act(() => {
-        const serverOutput = ReactDOMServer.renderToString(testHarness(props))
-        container.innerHTML = serverOutput
+        try {
+          const serverOutput = ReactDOMServer.renderToString(testHarness(props))
+          container.innerHTML = serverOutput
+        } catch (e: unknown) {
+          rendererProps.setError(e as Error)
+        }
```
mpeyper, line 28: turns out error boundaries throw synchronously when server rendering.  This is the only renderer that has any special treatment for error handling and only in the initial render function.  The error boundary behaves normally once it has been hydrated.

src/helpers/createTestHarness.tsx
```diff
@@ -1,42 +1,65 @@
...
+  // out users from having to work it out, we crudely suppress the output matching the patterns
+  // below.  For more information, see these issues:
+  //   - https://github.com/testing-library/react-hooks-testing-library/issues/50
+  //   - https://github.com/facebook/react/issues/11098#issuecomment-412682721
+  //   - https://github.com/facebook/react/issues/15520
+  //   - https://github.com/facebook/react/issues/18841
+  const removeConsoleFilter = filterConsole(
+    [
+      /^The above error occurred in the <TestComponent> component:/, // error boundary output
+      /^Error: Uncaught .+/ // jsdom output
```
mpeyper, line 20: I'm a little worried that this one is a bit _too_ generic, however, the suppression is removed in the test cleanup, so it should limit any potential smothering of useful logs that might be have been shown.

src/helpers/createTestHarness.tsx
```diff
@@ -1,42 +1,65 @@
...
-    } else {
-      setError(err as Error)
+function suppressErrorOutput() {
+  // The error output from error boundaries is notoriously difficult to suppress.  To save
+  // out users from having to work it out, we crudely suppress the output matching the patterns
+  // below.  For more information, see these issues:
+  //   - https://github.com/testing-library/react-hooks-testing-library/issues/50
+  //   - https://github.com/facebook/react/issues/11098#issuecomment-412682721
+  //   - https://github.com/facebook/react/issues/15520
+  //   - https://github.com/facebook/react/issues/18841
```
mpeyper, line 16: I wonder if we need all this information here or is it just noise to anyone else coming in after that change is made?

src/helpers/createTestHarness.tsx
```diff
@@ -1,42 +1,65 @@
...
-  rendererProps: RendererProps<TProps, TResult>,
+  { callback, setValue, setError }: RendererProps<TProps, TResult>,
   Wrapper?: WrapperComponent<TProps>,
   suspense: boolean = true
 ) {
+  const TestComponent = ({ hookProps }: { hookProps?: TProps }) => {
+    // coerce undefined into TProps, so it maintains the previous behaviour
+    setValue(callback(hookProps as TProps))
+    return null
+  }
```
mpeyper, line 38: I brought this into `createTestHarness` because:

1. `ErrorFallback` needed to be inside and I like consistency
2. It simplified the `props` of `TestComponent` as the `rendererProps` are already available in the outer scope

src/helpers/createTestHarness.tsx
```diff
@@ -1,42 +1,65 @@
...
+
+  let resetErrorBoundary = () => {}
+  const ErrorFallback = ({ error, resetErrorBoundary: reset }: FallbackProps) => {
+    resetErrorBoundary = () => {
+      resetErrorBoundary = () => {}
+      reset()
+    }
```
mpeyper, line 45: This is the clunky bit I was referring to.  We set a dummy reset as we don't get the real reset until the first error is called.  When doing the real reset, we set it back to the dummy reset in case the next time an error is thrown there is a new reset function provided (I've read the `react-error-boundary` code and that is not how it works, but fome an interface perspective we cannot be certain it will never change.

src/helpers/createTestHarness.tsx
joshuaellis: No, keep it here, this some of the issues you posted are still open so I imagine *if* they do something to solve these, then it'll have an impact, i like the paper trail 🔍

src/helpers/createTestHarness.tsx
joshuaellis: I think I originally moved it out before we moved out `testHarness` so It makes sense to put it back in since `TestComponent` isn't required on it's own.

src/helpers/createTestHarness.tsx
joshuaellis: I guess if we can't be certain we have to assume the worse.

joshuaellis (approved):
Looks good, be good to monitor over the lifecycle of react too, think we might need to watch carefully.

PR #541, opened 2021-01-15
feat: add option to not suppress error output if desired

mpeyper:
**What**:

Add `suppressErrorOutput` option (default to `true`) to allow disabling of the `console.error` suppression we added with the error boundary fix for `useEffect` errors.

**Why**:

It was [mentioned in discord](https://discord.com/channels/723559267868737556/723852521734668299/799444299870896169) that it might be causing issues for some people as its nuked their own console mocks.

**How**:

Just a `boolean` option an a `if` statement to skip over the call that uses `filter-console`.

**Checklist**:

- [ ] Documentation updated
- [x] Tests
- [ ] Ready to be merged
      

I've put this up for discussion to see if:

1. this is something we want to add and support
2. is there a better way we could have handled the error logs than this

Any feedback is welcome.

### DIFF
--- a/src/dom/__tests__/errorHook.test.ts
+++ b/src/dom/__tests__/errorHook.test.ts
@@ -142,4 +142,42 @@ describe('error hook tests', () => {
       expect(result.error).toBe(undefined)
     })
   })
+
+  describe('error output suppression', () => {
+    const originalConsoleError = console.error
+    const mockConsoleError = jest.fn()
+
+    beforeEach(() => {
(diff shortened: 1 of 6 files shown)


joshuaellis (changes requested):
As I mentioned in discord, I think this is probably the best way, I also definitely agree with default being true to supress. We should add it to the documentation and then it's good to go.

mpeyper:
Closing in favour of #549

PR #542, opened 2021-01-15
fix: reduce console.error suppressions to only while acting

mpeyper:
**What**:

Another option vs #541 

**Why**:

Making it _just work_ for more users without having to set options.

**How**:

By only suppressing the error boundary output while `act`ing, the test code retains access to the unaltered (or altered in their own way) version of `console.error`.

Technically, the test code will have access to our suppressed version of `console.error` within an `act` callback, however they would have to be using it for something other than logging an error (e.g. asserting a mocked version was called) within that callback to have an issue.  Calling it to log an error with it would work as expected.

There is an option to return the unaltered `act` to them and only use the wrapped version internally around `render` calls, which would alleviate the above issue, however that might leave them susceptible to showing the output if their `act` update caused and error in the subsequent render passes.

The tradeoff with this implementation is that we cannot pass an option to disable it like in #541 as the `act` function needs to be wrapped in the global scope before the the renderer receives the options.  We could potentially use an environment variable to disable is globally [like we do for other things](https://github.com/testing-library/react-hooks-testing-library/blob/master/dont-cleanup-after-each.js#L1) if we really want an escape hatch on it.

**Checklist**:

- [x] Tests
- [ ] Ready to be merged
      

The code here is pretty crude, especially in the `act` wrapper.  I'll clean it up if we want to go this way.

### DIFF
--- a/src/helpers/act.ts
+++ b/src/helpers/act.ts
@@ -1,0 +1,28 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+import { Act } from '../types/react'
+
+import { suppressErrorOutput } from './console'
+
+function createActWrapper(baseAct: Act) {
+  const act: Act = async (callback: () => any) => {
+    const restoreOutput = suppressErrorOutput()
+    try {
+      let awaitRequired = false
+      const actResult = (baseAct(() => {
+        const callbackResult = callback()
+        awaitRequired = callbackResult !== undefined && !!callbackResult.then
+        return callbackResult
+      }) as any) as PromiseLike<undefined>
+
+      return awaitRequired ? await actResult : undefined
+    } finally {
+      restoreOutput()
+    }
+  }
+
+  return act
+}
+
+export { createActWrapper }
--- a/src/helpers/act.ts
+++ b/src/helpers/act.ts
@@ -1,21 +1,19 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
-/* eslint-disable @typescript-eslint/no-unsafe-assignment */
-/* eslint-disable @typescript-eslint/no-unsafe-member-access */
 import { Act } from '../types/react'
 
 import { suppressErrorOutput } from './console'
 
+import { isPromise } from './promises'
+
 function createActWrapper(baseAct: Act) {
-  const act: Act = async (callback: () => any) => {
+  const act: Act = async (callback: () => unknown) => {
     const restoreOutput = suppressErrorOutput()
     try {
       let awaitRequired = false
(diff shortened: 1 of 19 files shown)


src/helpers/act.ts
```diff
@@ -0,0 +1,28 @@
...
+      }) as any) as PromiseLike<undefined>
+
+      return awaitRequired ? await actResult : undefined
+    } finally {
+      restoreOutput()
+    }
+  }
+
+  return act
+}
```
mpeyper, line 26: I am not proud of this.  Any ideas for tidying it up are welcome (especially all the `eslint` disables)

joshuaellis: I've made some suggestions in a seperate branch as It was easier, i'll PR it to this branch now 👍🏼

joshuaellis:
I'm struggling to understand though why we'd want to impliment this "just working" feat as opposed to getting users to add one param? This seems like overall it's more effort to maintain?

mpeyper:
I agree that it's more to maintain and we should be wary of that.

The main reason to consider over #541 is that by limiting the side effect (replacing `console.error`), it should work out-of-the-box for most, if not all, users. 

In #541, setting the option to `false` gives them back the original `console.error` but they then also have to handle the react output, whether that be suppressing it themselves or additional calls they need to consider in their expectations. In this change they can use ` console.error` like before and the react output is filtered out if and doesn't pollute their expectations.

I'm starting to think an environment variable escape hatch is worth it if we go this way. I'm still very much on the fence about it though as that act wrapper scares me a bit.

joshuaellis:
Maybe let's release a beta, get the folks on discord to try it out & see how they feel – RFC. We could also add `@latest` to the install doc to coerce some into using the beta? But an escape hatch would also be good too I think. 

RE: Act wrapper, my concern is they change it one impl in either `react-dom` or `react-test-renderer` and then we have to create more than one wrapper... But we can always see how it goes. It seems this is a good starting point to trial?

joshuaellis (approved):
Comments above

joshuaellis:
That’s funny, I was about to pick this up, so do you think this should be the route to go with? Whilst incl. an ENV escape hatch?

mpeyper:
Even though the code here is a bit more hectic than #541, I feel like this would be the easiest to unwind later of we want to because it doesn't introduce a new option (other than the env variable)

mpeyper:
Feels like everyone else is saying go for #541 though 😕

joshuaellis:
Wouldn’t the variable work the same as the option? Therefore this PR is 541 but a smarter(?)

mpeyper:
> Wouldn’t the variable work the same as the option

Similar but not quite the same.  The env variable is much more global than the option which is constrained to a single test.  In practice though you'd probably get every test in a file passing the option and it might as well be global at that point.

mpeyper:
Closing in favour of #549

PR #543, opened 2021-01-15
refactor: reduce eslint-warnings & type changes

joshuaellis:
I think we removed `isPromise` thinking we didn't need it, but it rears it's head again! I also added the undefined bit just to make the act wrapper function a bit cleaner.


mpeyper:
This is better, thanks.

PR #549, opened 2021-01-18
fix: only suppress console.error for non-pure imports

### LINKED ISSUE #546: `renderHook` from /dom/pure has global side-effects
- `react-hooks-testing-library` version: 5.0.0
- `react-test-renderer` version: not installed
- `react` version: 17.0.1
- `node` version: 17.0.1
- `npm` (or `yarn`) version: `yarn@1.22.0`

### Relevant code or config:

```js
const { renderHook } = require("@testing-library/react-hooks/dom/pure");

let consoleError;
beforeEach(() => {
  consoleError = console.error;
});

afterEach(() => {
  if (console.error !== consoleError) {
    throw new Error("Did not cleanup console.error");
  }
});

test("pure", () => {
  function useHook() {
    return useState(false);
  }

  const { result } = renderHook(() => useHook(), {});
});
```

### What you did:

Use `@testing-library/dom/pure` in `karma-mocha` though the problem happens in arbitrary testing frameworks I suspect

### What happened:

`console.error` was not cleaned up

### Reproduction:

https://github.com/eps1lon/testing-library-react-hooks-impure

### Problem description:

We check for mistakes in `afterEach` if a test was not properly cleaned up. One of those checks looks at `console.error`. 

### Suggested solution:

`renderHook` probably has some global side-effects. These should be documented by explaining why they exists and how to disable them or work around them.

mpeyper:
**What**:

Yet another alternative to #541 and #542

**Why**:

#546

**How**:

The complete opposite approach to #542 but the documentation and tests are still valid.  By extending the suppression time from the initial import to after all tests have run, the patched `console.error` can be used by other utilities when setting up local mocks, while the code remains relatively straight forward and we don't need to wrap `act`.

Care here is taken to ensure that patching only occurs if the defaul imports are used and not the `/pure` imports.

**Checklist**:

- [x] Documentation updated
- [x] Tests
- [x] Ready to be merged
      

Fixes #546


mpeyper:
I wasn't quite right about the docs and tests not needing updating.  The docs were _very_ close, but I have added more detail about the side effects, and the tests were still accurate, but there were some new cases that needed a test so I have refactored them a bit too.

joshuaellis:
So would the way to disable the console patching to be to use the `/pure` export of each module, else we continue to do what we're currently doing?

mpeyper:
Using the pure import will disable it for the test file or setting the env variable (e.g. with the setup file utility) can disable it more globally.  [The docs cover these options](https://github.com/testing-library/react-hooks-testing-library/blob/698579c1c23f05557979f873293624aa6f5b3e42/docs/api-reference.md#consoleerror).

docs/api-reference.md
```diff
@@ -12,6 +12,7 @@ route: '/reference/api'
 - [`cleanup`](/reference/api#cleanup)
 - [`addCleanup`](/reference/api#addcleanup)
 - [`removeCleanup`](/reference/api#removecleanup)
+- [`console.error`](/reference/api#consoleerror)
```
mpeyper, line 15: Just reading what this list of links is for, I don't this this one should be added here after all (it will appear in the side menu still though).

mpeyper:
Honestly, I'm feeling best about this option at the moment, especially now the suppression is being handled in before/after test hooks.

mpeyper:
> especially now the suppression is being handled in before/after test hooks.

Now that I've just said that I wonder if the "on import" of the previous change makes it slightly better for people wanting to mock `console.error` so that their mock is mock likely to occur after our patching.

Or moving to to `beforeAll` and `afterAll` instead of `...Each` could also make sense as they are more likely to be using `...Each` (not may people is `...All` in my experience, but that might just be me).

... Back to confusion.

joshuaellis:
I've had a quick look at this, it does look promising. Resolving the `pure` import was a good step to take & i think taking advantage of `*Each` functions in Jest was a good idea too. I'm especially happy we don't seem to be wrapping `act` anymore?

mpeyper:
>  I'm especially happy we don't seem to be wrapping act anymore

Correct.  No `act` wrapping.

I think we should move forward with this one.  It's not worse than what we have now, but also has the escape hatches if people need them.

joshuaellis (approved):
Looks good, well done for exploring so many options, I don't think I would have been able to tackle this issue so thoroughly 😅

mpeyper:
Ok... **deep breath**... here we go...

Issue #564, opened 2021-02-19
Noisy React console error output is not automatically suppressed as documented when using test frameworks other than Jest


jaydenseric:
- `react-hooks-testing-library` version: v5.0.3

### Reproduction:

In a blank project, run:

```
npm install react-test-renderer @testing-library/react-hooks --save-dev
```

In `test.js`:

```js
const { renderHook } = require('@testing-library/react-hooks');

function useExample() {
  throw new Error('Message.');
}

renderHook(() => useExample());
```

Run:

```
node test.js
```

Notice the unexpected console error output in the terminal:

<img width="1223" alt="Screen Shot 2021-02-19 at 11 13 14 am" src="https://user-images.githubusercontent.com/1754873/108438555-9e89d180-72a3-11eb-8db9-3867ced67118.png">

```
The above error occurred in the <TestComponent> component:

    at TestComponent (/[redacted]/node_modules/@testing-library/react-hooks/lib/helpers/createTestHarness.js:20:5)
    at Suspense
    at ErrorBoundary (/[redacted]/node_modules/react-error-boundary/dist/react-error-boundary.cjs.js:40:35)

React will try to recreate this component tree from scratch using the error boundary you provided, ErrorBoundary.
```

### Problem description:

The documentation says:

> To keep test output clean, we patch `console.error` when importing from `@testing-library/react-hooks` (or any of the other non-pure imports) to filter out the unnecessary logging and restore the original version during cleanup.
> — https://github.com/testing-library/react-hooks-testing-library/blob/master/docs/api-reference.md#consoleerror

In truth, patching only occurs if certain `beforeEach` and `afterEach` globals, arbitrary to [specific testing frameworks](https://jestjs.io/docs/en/api#aftereachfn-timeout), are present:

https://github.com/testing-library/react-hooks-testing-library/blob/c6de60e5a04d49d345ee6680fa5c02c75c702644/src/core/console.ts#L6-L7

Presumably it's assumed that every project in the world uses Jest 🙃

In another part of the docs:

> In order to run tests, you will probably want to be using a test framework. If you have not already got one, we recommend using Jest, but this library should work without issues with any of the alternatives.
> — https://github.com/testing-library/react-hooks-testing-library/blob/master/docs/installation.md#testing-framework

[...]

PR #571, opened 2021-03-01
feat(console): export helper to manually suppress error output

mpeyper:
**What**:

Add a helper to suppress the error output in the same way the automatic suppression does for users that can't use the automatic version.

Also extended the documentation around this feature to describe the requirements for automatic suppression and give instructions on how to do it manually.

**Why**:

#564 

**How**:

extracted the logic from the `beforeEach` and `afterEach` and exported them from all the modules.

**Checklist**:

- [x] Documentation updated
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/dom/__tests__/errorSuppression.disabled.test.ts
+++ b/src/dom/__tests__/errorSuppression.disabled.test.ts
@@ -5,10 +5,10 @@ describe('error output suppression (disabled) tests', () => {
 
   beforeAll(() => {
     process.env.RHTL_DISABLE_ERROR_FILTERING = 'true'
+    require('..')
   })
 
   test('should not patch console.error', () => {
-    require('..')
     expect(console.error).toBe(originalConsoleError)
   })
 })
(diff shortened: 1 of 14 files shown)


src/dom/__tests__/errorSuppression.disabled.test.ts
```diff
@@ -5,10 +5,10 @@ describe('error output suppression (disabled) tests', () => {
 
   beforeAll(() => {
     process.env.RHTL_DISABLE_ERROR_FILTERING = 'true'
+    require('..')
   })
 
   test('should not patch console.error', () => {
-    require('..')
```
mpeyper, line 11: This actually wasn't testing anything.  Requiring in the test wouldn't register the `beforeEach` and `afterEach` so it must be done before the test executes.  I discovered this because with the refactoring, this left a uncovered branch that is now covered.

joshuaellis (approved):
Looks great, documentation is clear as well 👍🏼

PR #593, opened 2021-04-10
chore: update dependabot.yml

MichaelDeBoey:
Also use Dependabot to check for actions updates


mpeyper:
Thanks!

@all-contributors please add @MichaelDeBoey for maintenance

PR #599, opened 2021-04-10
chore: update validate.yml

MichaelDeBoey:


joshuaellis:
Would you mind explaining for someone out of the loop, why we need to make this change please? 😄

MichaelDeBoey:
@joshuaellis
- I've added `node@v10`, because otherwise the tests won't run against latest v10, only against v10.13 specific
- `access_token` isn't required anymore by [`styfle/cancel-workflow-action`](https://github.com/styfle/cancel-workflow-action)
- Removed the specific version of [`actions/setup-node`](https://github.com/actions/setup-node), so Dependabot won't bug you with every single patch update

mpeyper:
What is the significance of node 10.13? I didn't set up the CI config and I believe it was modelled off what `react-testing-library` was doing at the time, but I've always been curious. Also, why do we skip node 13?

MichaelDeBoey:
> What is the significance of node 10.13?

@mpeyper `10.13` is the lowest supported Node version

> Also, why do we skip node 13?

All odd versions are considered unstable, so we're only testing against latest (`15`) to check whether we would break on newest version or not

mpeyper:
Sounds good.

Are you sure about the cancel action not needing a token anymore? [Their docs](https://github.com/styfle/cancel-workflow-action#usage) still mention it?

MichaelDeBoey:
@mpeyper I'm sure
https://github.com/styfle/cancel-workflow-action/releases/tag/0.7.0

mpeyper:
I feel like I'm missing something with how the cancel action works. By reading [the code](https://github.com/styfle/cancel-workflow-action/blob/0.8.0/src/index.ts#L27) it looks like `access_token` is a required "input", but I can see in the action logs that it's logging it found the token, but I have NFI how it found it.

I guess I don't really need to know either, but I am curious now 😅

mpeyper:
Ok, all looks good. Thanks again 😃

PR #604, opened 2021-04-21
chore: add tests for Node 16

MichaelDeBoey:
https://nodejs.medium.com/node-js-16-available-now-7f5099a97e70


joshuaellis:
We've removed node 15 from the list but it's passing is required. We'll need to sort this out before merging.

mpeyper:
I've updated the required checks to remove 15 and add 16.

Issue #605, opened 2021-04-21
Usage in close-to-real SSR environment?


xobotyi:
Is there any way to test SSR functionality in environment close to real SSR?

I'm testing the hook that should fallback to noop during SSR, the hook declaration itself, as for now i'm checking SSR by testing `window` type: `typeof window !== 'undefined'`. Like this:
```ts
export const myAwesomeHook = typeof window === 'undefined' ? ()=>{} : ()=>{/*some real hook code*/}
```

When testing with jest in it's common env (jsdom) window object is obviously exists and there is no way to test that everything will be fine in SSR, since i gaij browser version of a hook.
If to switch environment to `node`, ssr tests failing with
```
    The error below may be caused by using the wrong test environment, see https://jestjs.io/docs/en/configuration#testenvironment-string.
    Consider using the "jsdom" test environment.

    ReferenceError: document is not defined

       7 |
    >  8 |     renderHook(() => myAwesomeHook());
         |     ^
       9 |

```

In this case renderHook improted from server: `import { renderHook } from '@testing-library/react-hooks/server';`

mpeyper:
Hi @xobotyi, thanks for raising this issue.  I think it's being caused by this line:

https://github.com/testing-library/react-hooks-testing-library/blob/ef1d731c2a0d7d78e3b33b183d5b5dd927785aae/src/server/pure.ts#L16

We should move that into the `hydrate` function so we reference `document` when they want it to render for real.

Are you interested in submitting a PR for this @xobotyi?

xobotyi:
@mpeyper i'm not really into SSR render flow, but if solution is only about that -i'll do.

PR #607, opened 2021-04-22
Add ability to run SSR tests in pure-node environment

xobotyi:
**What**:
Fix: #605

**Why**:
Before that change it was impossible to run SSR test in pure-nod eenvironment, where no DOM global objects exists.

**How**:
Div element now created only in hydration stage (also now it is a marker of hydration occurred)

**Checklist**:
- [ ] Documentation updated
- [ ] Tests
- [x] Ready to be merged

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -519,6 +519,16 @@
       "contributions": [
         "maintenance"
       ]
+    },
+    {
+      "login": "xobotyi",
+      "name": "Anton Zinovyev",
+      "avatar_url": "https://avatars.githubusercontent.com/u/6178739?v=4",
+      "profile": "https://github.com/xobotyi",
+      "contributions": [
+        "bug",
+        "code"
+      ]
     }
   ],
   "skipCi": true,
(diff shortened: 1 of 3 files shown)


mpeyper (approved):
This looks good. Have you checked if it resolves the issue in your test?

It would be nice if we could have sone server only tests run in the `node` environment without `jsdom` to make sure we don't break it in the future, but I'm not sure how we would do that for just some tests? I won't hold this up for it if it too hard though.

xobotyi:
@mpeyper yest, i've monkeypatched local version - it solves the issue.

About server tests - it is easy with recent jest projects feature (the best they've introduced yet IMO), but here used another test runner so i dont know how to do it here.
As you can see below all my hooks are tested bont against `node` and `jsdom` environments on the single run:
image

mpeyper:
Oh damn, that looks sweet. I'll definitely be investigating that to clean up our bloated test suite.

Thanks for the fix, I'll merge this now. New version should be out soon for you (bot will comment here).

PR #622, opened 2021-05-17
fix(types) Correct overload order for act without strictNullChecks in TypeScript

andyrooger:
**What**:

Correct the TypeScript definitions, so that we produce the correct return type for `act(...)` calls when `strictNullChecks` is turned off.

**Why**:

Take the following code

```ts
import { act } from '@testing-library/react-hooks'

const result = act(() => Promise.resolve())
```
With `strictNullChecks` turned on, the type of `result` is `Promise<undefined>` as you would expect.
Turn `strictNullChecks` off, and the type of `result` is now `void`.

This is particularly highlighted in my case as the project I'm working on has eslint rules that will complain if we try to await a value it thinks is not thenable.

**How**:

It seems the issue is that with `strictNullChecks`, the two overloads for `act(...)` do not have a lot of overlap so the order we specify them in doesn't matter. Without `strictNullChecks`, anything that matches the async overload also matches the sync overload. Since the sync overload is specified first, it is always chosen.

I've just swapped the overloads around so that the more specific one with Promises is checked first.

**Checklist**:

- [x] Tests
- [x] Ready to be merged
      
### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -547,6 +547,15 @@
       "contributions": [
         "maintenance"
       ]
+    },
+    {
+      "login": "andyrooger",
+      "name": "andyrooger",
+      "avatar_url": "https://avatars.githubusercontent.com/u/420834?v=4",
+      "profile": "https://github.com/andyrooger",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "skipCi": true,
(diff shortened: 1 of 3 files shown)


andyrooger:
@mpeyper I've followed the contributors doc for this, but it looks like you're using a bot for adding contributors now. Do I need to strip that change out?

joshuaellis:
No you don't, the bot is just to help us if someone forgets!

mpeyper:
Thanks @andyrooger, I can see [there is actually a note about this very issue](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-dom/test-utils/index.d.ts#L292-L294) in react's own types for `act`.

andyrooger:
@mpeyper so they do! That could have saved me the long explanation. Thanks for the merge :)

PR #623, opened 2021-05-18
chore: drop Node 10 support

### LINKED ISSUE #621: Drop Node 10 support
Because it's [EOL](https://nodejs.org/en/about/releases).

mpeyper:
**What**:

Drop node 10 support

**Why**:

Fixes #621 

**How**:

Remove node `10.13` and `10` from node version matrix in github actions.

**Checklist**:

- [x] Ready to be merged

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -16,7 +16,7 @@ jobs:
     if: ${{ !contains(github.head_ref, 'all-contributors') }}
     strategy:
       matrix:
-        node: [10.13, 10, 12, 14, 16]
+        node: [12.13, 12, 14, 16]
     runs-on: ubuntu-latest
     steps:
       - name: 🛑 Cancel Previous Runs
@@ -48,8 +48,8 @@ jobs:
     runs-on: ubuntu-latest
     if:
       ${{ github.repository == 'testing-library/react-hooks-testing-library' &&
-      contains('refs/heads/main,refs/heads/beta,refs/heads/next,refs/heads/alpha',
-      github.ref) && github.event_name == 'push' }}
+      contains('refs/heads/main,refs/heads/beta,refs/heads/next,refs/heads/alpha', github.ref) &&
+      github.event_name == 'push' }}
     steps:
       - name: 🛑 Cancel Previous Runs
         uses: styfle/cancel-workflow-action@0.9.0
--- a/package.json
+++ b/package.json
@@ -4,6 +4,9 @@
   "description": "Simple and complete React hooks testing utilities that encourage good testing practices.",
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
+  "engines": {
+    "node": ">=12"
+  },
   "keywords": [
     "testing",
     "react",


.github/workflows/validate.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     if: ${{ !contains(github.head_ref, 'all-contributors') }}
     strategy:
       matrix:
-        node: [10.13, 10, 12, 14, 16]
+        node: [12, 14, 16]
```
mpeyper, line 19: Is there be a minimum version of v12 we should be testing against?

.github/workflows/validate.yml
```diff
@@ -48,8 +48,8 @@ jobs:
     runs-on: ubuntu-latest
     if:
       ${{ github.repository == 'testing-library/react-hooks-testing-library' &&
-      contains('refs/heads/main,refs/heads/beta,refs/heads/next,refs/heads/alpha',
-      github.ref) && github.event_name == 'push' }}
+      contains('refs/heads/main,refs/heads/beta,refs/heads/next,refs/heads/alpha', github.ref) &&
+      github.event_name == 'push' }}
```
mpeyper, line 52: I think this was an auto format.

.github/workflows/validate.yml
MichaelDeBoey: `12.13` was first LTS version

https://nodejs.org/en/blog/release/v12.13.0

```suggestion
        node: [12.13, 12, 14, 16]
```

MichaelDeBoey: But we can also check against Node 12.0 if we want

```suggestion
        node: ['12.0', 12, 14, 16]
```

MichaelDeBoey (approved):

joshuaellis (approved):

package.json
```diff
@@ -4,6 +4,9 @@
   "description": "Simple and complete React hooks testing utilities that encourage good testing practices.",
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
+  "engines": {
+    "node": ">=12"
+  },
```
mpeyper, line 9: I figured if we were going to make a breaking change, we should at least have something breaking in there 😅

mpeyper:
Can somebody just rereview this and check the [commit message](https://github.com/testing-library/react-hooks-testing-library/pull/623/commits/da3dedb814643ac5c4254a9acfd07fd8b043e488) is correct for a major bump.  I always seem to mess up the order of the breaking changes, fixes and co-authors an either the wrong version gets published or the release notes get all messed up.

joshuaellis:
It _should_ cause a breaking change release because it's in the body. I couldn't say about release notes, they're still a mystery to me.

mpeyper:
I found [this](https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit-message-footer) which was linked from the semantic-release docs which suggests I need 2 blank lines after the breaking change summary otherwise it will think it is the breaking change description...  I'll put an extra one in to be safe 🤷‍♂️

joshuaellis (approved):
Yeah the breaking change system always gets me :(

mpeyper:
Annoyingly, git is trimming the double blank line into a single new line when I commit... oh well, guess I'll just have to double check the release notes after it runs.

joshuaellis:
I didn't double line break yesterday and it worked for me if incl release notes...

PR #624, opened 2021-05-21
feat: removed filter-console dependency and fallback if process.env is not available

### LINKED ISSUE #617: use of process.env and filter-console library prevents running tests in browser environment
- `react-hooks-testing-library` version: 5.1.2
- `react` version: 17
- `react-dom` version (if applicable): 17
- `react-test-renderer` version (if applicable): n/a
- `node` version: 14
- `npm` (or `yarn`) version: yarn 2

### Relevant code or config:

`import '@testing-library/react-hooks/dom/pure';`

### What you did:

just imported the library

### What happened:

```
ERROR in ./node_modules/filter-console/index.js 2:13-28
Module not found: Error: Can't resolve 'util' in '/Users/bwain/dev/react-hooks-testing-lib-process-bug/node_modules/filter-console'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.
```

I've also seen other errors (depending on my exact config) due to use of `process.env` for things like `process.env.RHTL_SKIP_AUTO_CLEANUP`.

### Reproduction:

https://github.com/bdwain/react-hooks-testing-lib-process-bug

### Problem description:

react testing library can not run in a browser environment because it uses `process.env` and because it pulls in the `filter-console` library, which is targeted at `node` and uses the `util` library under the hood. 

Webpack used to provide a node polyfill for you automatically in webpack <5, which likely masked the issue for a long time. Now that they've stopped polyfilling node, issues like this are being exposed.

While Jest is probably the most widely use test runner right now (and it uses a node environment with JSDom), many people prefer to run their tests in an actual browser using tools like [karma](http://karma-runner.github.io/latest/index.html). There's no good way to do that though with the current usages of `process.env` and `filter-console`. If we were to start modifying our test environments to polyfill node modules or stub them out somehow, we'd lose our ability to detect usages of those node modules in tests, which decreases their value.

### Suggested solution:

If the usages of node modules could be replaced with things that work in both a browser environment and a node environment, that would open up usage to a lot of people.

Alternatively, the things that use node modules could be made to be separately imported from the rest of the library, so that they are opt-in rather than forced on you.

mpeyper:
**What**:

Remove Node specific dependencies to make running tests in a non-node environment (i.e. the browser) is easier to achieve without defining variables or polyfilling APIs.

**Why**:

Fixes #617

**How**:

Removed `filter-console` dependency as it was dependant on Node's `util` module.

Also wrapped access of `process.env` variables to catch if undefined and return default values.

**Checklist**:

- [x] Tests
- [x] Ready to be merged
      

There is a potentially breaking change here, depending on how you view it.  Previously, the `process.env.RHTL_DISABLE_ERROR_FILTERING` check was performed inside the `suppressErrorOutput` function, which was fine when it was first introduced.

Later, `suppressErrorOutput` was exposed out of the API so that consumers using the `pure` imports could opt into the suppression as well, however, setting the `RHTL_DISABLE_ERROR_FILTERING` variable had the possibility of making explicit calls to `suppressErrorOutput` a no-op.

I view this as a bug introduced when exposing `suppressErrorOutput`, rather than a breaking change being introduced here now, and I doubt anyone is relying on this odd combination of behaviour, but if others feel strongly I'm happy to either retain the existing behaviour or make this a major bump.  Feedback welcome.

Finally, I've snuck in a few pieces of cleanup from dropping Node 10 support as well as some small tweaks that makes the final output of the code not rely on `@babel/runtime` at all.  I'm loathed to remove `@babel/runtime` dependency all together as I feel it's too easy for a change to introduce the need for it again, which is relatively unknown until you publish a broken version.


src/core/console.ts
```diff
@@ -1,24 +1,40 @@
-import filterConsole from 'filter-console'
+const consoleFilters = [
+  /^The above error occurred in the <.*?> component:/, // error boundary output
+  /^Error: Uncaught .+/ // jsdom output
+]
 
 function suppressErrorOutput() {
-  if (process.env.RHTL_DISABLE_ERROR_FILTERING) {
-    return () => {}
-  }
```
mpeyper, line 6: This is the breaking change I referred to in the PR description

src/core/console.ts
```diff
@@ -1,24 +1,40 @@
-import filterConsole from 'filter-console'
+const consoleFilters = [
+  /^The above error occurred in the <.*?> component:/, // error boundary output
```
mpeyper, line 2: I've slightly loosened what this will match, in case a `Wrapper` throws instead of the `TestComponent`

src/core/console.ts
```diff
@@ -1,24 +1,40 @@
...
-  return filterConsole(
-    [
-      /^The above error occurred in the <TestComponent> component:/, // error boundary output
-      /^Error: Uncaught .+/ // jsdom output
-    ],
-    {
-      methods: ['error']
+  const error = (...args: Parameters<typeof originalError>) => {
+    const message = typeof args[0] === 'string' ? args[0] : null
+    if (!message || !consoleFilters.some((filter) => filter.test(message))) {
```
mpeyper, line 11: Just to call it out, `filter-console` would format the string before testing it against the patter (that's why it had a dependency on `util`), but I found this wasn't necessary for the cases we are trying to filter.  Not doing this may be an issue if how the calls from upstream are made ever change or if we introduce any new cases in the future (although I don't think that is likely).

src/dom/__tests__/autoCleanup.noProcessEnv.test.ts
```diff
@@ -0,0 +1,40 @@
...
+  let cleanupCalled = false
+  let renderHook: ReactHooksRenderer['renderHook']
+
+  beforeAll(() => {
+    process.env = {
+      ...process.env,
+      get RHTL_SKIP_AUTO_CLEANUP(): string | undefined {
+        throw new Error('expected')
+      }
+    }
```
mpeyper, line 18: I couldn't remove `process.env` from the test environment all together (other things started to crash), but I could make it error when trying to access the variables.

src/dom/__tests__/errorHook.test.ts
```diff
@@ -142,51 +142,4 @@ describe('error hook tests', () => {
       expect(result.error).toBe(undefined)
     })
   })
-
-  describe('error output suppression', () => {
```
mpeyper, line 146: this moved to [a new file](https://github.com/testing-library/react-hooks-testing-library/pull/624/files#diff-8bbf97dce934ce64ecc78bc41856f39d3fbdeca3cfd48ab25f21555649e1a677)

src/server/__tests__/autoCleanup.disabled.test.ts
```diff
@@ -1,30 +1,38 @@
...
 // This verifies that if RHTL_SKIP_AUTO_CLEANUP is set
 // then we DON'T auto-wire up the afterEach for folks
 describe('skip auto cleanup (disabled) tests', () => {
-  let cleanupCalled = false
-  let renderHook: ReactHooksRenderer['renderHook']
+  const cleanups: Record<string, boolean> = {
+    ssr: false,
+    hydrated: false
+  }
+  let renderHook: ReactHooksServerRenderer['renderHook']
```
mpeyper, line 12: I've improved quite a few of the server renderer tests to actually test the behaviour pre/post hydration.  Some already did this but others not so I made them consistent.

src/dom/pure.ts
```diff
@@ -1,4 +1,4 @@
-import ReactDOM from 'react-dom'
+import * as ReactDOM from 'react-dom'
```
mpeyper, line 1: This prevents the import relying on `@babel/runtime` for default import interop.

src/pure.ts
```diff
@@ -9,7 +9,7 @@ function hasDependency(name: string) {
   try {
     require(name)
     return true
-  } catch (error) {
+  } catch {
```
mpeyper, line 12: No longer needed once Node 10 was dropped (actually produced a lint warning in my local environment).

src/types/index.ts
```diff
@@ -59,6 +63,12 @@ export type RenderHookResult<
   Omit<TRenderer, keyof Renderer<TProps>> &
   AsyncUtils
 
+export type ServerRenderHookResult<
+  TProps,
+  TValue,
+  TRenderer extends ServerRenderer<TProps> = ServerRenderer<TProps>
+> = RenderHookResult<TProps, TValue, TRenderer>
```
mpeyper, line 70: I introduced a few server render specific types here to help out in the tests.  The technically aren't used by the production code, but they don't hurt it either.

src/server/pure.ts
```diff
@@ -1,5 +1,5 @@
-import ReactDOMServer from 'react-dom/server'
-import ReactDOM from 'react-dom'
+import * as ReactDOMServer from 'react-dom/server'
```
bdwain, line 1: shouldn't this require a change to how you use these also?

src/core/console.ts
```diff
@@ -1,24 +1,40 @@
...
+  console.error = error
+
+  return () => {
+    console.error = originalError
+  }
+}
+
+function errorFilteringDisabled() {
+  try {
+    return !!process.env.RHTL_DISABLE_ERROR_FILTERING
```
bdwain, line 25: might be worth putting a comment here explaining why try/catch is needed.

src/core/cleanup.ts
```diff
@@ -18,9 +18,17 @@ function removeCleanup(callback: CleanupCallback) {
   cleanupCallbacks = cleanupCallbacks.filter((cb) => cb !== callback)
 }
 
+function skipAutoCleanup() {
+  try {
+    return !!process.env.RHTL_SKIP_AUTO_CLEANUP
```
bdwain, line 23: same here with the commment

src/server/pure.ts
bdwain: i see in the other pure file

src/dom/pure.ts
bdwain: so this works the same either way? interesting

bdwain:
there are uses of `process.env` in dont-cleanup-after-each.js and disable-error-filtering.js too. Maybe wrap those in a try/catch that throws a helpful error output or something?

bdwain (approved):

bdwain:
this looks great! Thanks!

src/dom/pure.ts
mpeyper: It's interesting because `import ReactDOM` and `import * as ReactDOM` are _not_ equivalent as per the ES6 import spec, but most compilers/bundlers treat them interchangeably.  Babel does this by inserting the default import interop helper when there is no default export for the module, like in [React](https://github.com/facebook/react/blob/master/packages/react/index.js)/[ReactDOM](https://github.com/facebook/react/blob/master/packages/react-dom/index.js)'s case.  By changing the the `import * as ...`, it's actually _more_ correct for these libraries.

mpeyper:
> there are uses of process.env in dont-cleanup-after-each.js and disable-error-filtering.js too. Maybe wrap those in a try/catch that throws a helpful error or something?

Good point.  They might be used in setup files and get bundled in too, so we should protect them in a similar same way.

mpeyper:
I'm getting cold feet about not pushing it as a [breaking change](https://github.com/testing-library/react-hooks-testing-library/pull/624#discussion_r636608276).  @joshuaellis what are your thoughts?

joshuaellis (approved):
I think this is fair to assume it's a fix, if I understand correctly, it's all about `process.env` not breaking in the browser but using that `env` variable will still stop the call, so in _theory_ everyone's tests should run the same. 

If someone's done a hack // workaround to solve this themselves then there's a cause it'll break, but you can hardly account for that.

mpeyper:
Not quite @joshuaellis. Currently, if someone set `RHTL_DISABLE_ERROR_FILTERING` to `true` and then imported and called `suppressErrorOutput` it would skip the suppression and the error logs would be produced.

With this changes, explicit calls to `suppressErrorOutput` would override the env variable and suppress the logs.

joshuaellis:
Aha, thank you for clearing that up. Then it still sounds like it could be a breaking change, albeit not exactly dramatic, but given previous history we've had with "breaking people's test suites" I think we should play it safe and release as breaking.

mpeyper:
Ok, here we go...

bdwain:
thanks @mpeyper

mpeyper:
@all-contributors please add @bdwain for bug, review 😄

PR #637, opened 2021-06-18
Prove wait for works  using useFakeTimers

### LINKED ISSUE #631: `waitFor` doesn't work if jest fake timers are used
- `react-hooks-testing-library` version: 7.0.0
- `react` version: 17.0.2
- `react-dom` version: 17.0.2
- `node` version: 14.16.0
- `npm` version: 7.10.0

### Problem
When using `waitFor` when Jest has been configured to use fake timers then the `waitFor` will not work and only "polls" once. After that the test just hangs until Jest comes in and fails the test with that the test exceeds the timeout time. Below is some code that showcases the problem.

```js

it('does not work', async () => {
  jest.useFakeTimers()

  const { waitFor } = renderHook(() => {})

  await waitFor(() => {
    console.log('poll') // Is printed just once
    expect(false).toBe(true)
  }, { timeout: 25, interval: 10 })

  // Fails with Exceeded timeout of 5000 ms for a test.
})
```

Basically the `waitFor` from `@testing-library/react-hooks` is using the faked `setTimeout` or `setInterval` which prevents it from working correctly.

There is a workaround (see suggested solution) but I recommend providing a nice error message when `waitFor` is used together with faked timers or maybe change the implemenation so it will work with fake timers.

### Suggested solution
I found this [issue](https://github.com/testing-library/dom-testing-library/issues/661) and it seems that person has already been fixed in `@testing-library/dom`. From my perspective I can suggest maybe reuse that function instead of implementing it yourselves but I don't really know the internal structure / code.

But after finding that issue and realizing that is has been fixed there, then I use the following code as a workaround which works fine.

```js
import { waitFor } from '@testing-library/react'

it('works', async () => {
  jest.useFakeTimers()

  await waitFor(() => {
    console.log('poll') // Is printed twice
    expect(false).toBe(true)
  }, { timeout: 25, interval: 10 })

  // Fails with false is not equal to true
})
```

### A more real world scenario
If curios on the actual problem I'm facing is to test the following hook:

```js
function useSomething({ onSuccess }) {
  const poll = useCallback(async () => {
    const result = await fetch(/* ... */)
    if (result.ok) onSuccess()
  }, [onSuccess])

  useEffect(() => {
    const id = setInterval(() => { poll() }, 2000)
    return () => clearInterval(id)
  }, [poll])
}
```

What I want to do is test that it invokes the `onSuccess` function on a successfull poll.

```js
it('invokes the `onSuccess` on successfull poll', async () => {
  const onSuccess = jest.fn()
  jest.useFakeTimers()

  const { waitFor } = renderHook(() => useSomething({ onSuccess }))

  jest.runOnlyPendingTimers()
  await waitFor(() => expect(onSuccess).toHaveBeenCalled())
})
```

chris110408:
**What**:

Fixes #631

**Why**:
Create the unit test to prove the wait for function works using jest.useFakeTimers

**How**:

Created a unit test use jest.useFakeTimers and jest.advanceTimersByTime()
**Checklist**:

- [ ] Documentation updated
- [ ] Tests
- [ ] Ready to be merged
      
### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "docz": "2.3.1",
     "docz-theme-default": "1.2.0",
     "docz-utils": "2.3.0",
-    "eslint": "7.28.0",
+    "eslint": "7.29.0",
     "kcd-scripts": "11.1.0",
     "prettier": "^2.2.1",
     "react": "17.0.2",


mpeyper:
thanks again @chris110408!

I'll wait for a resolution to the [conversation in #631](https://github.com/testing-library/react-hooks-testing-library/issues/631#issuecomment-863710645) before doing anything with this.

src/dom/__tests__/asyncHook.test.ts
```diff
@@ -100,6 +100,54 @@ describe('async hook tests', () => {
...
+    await waitFor(
+      () => {
+        expect(actual).toBe(expected)
+        complete = true
+      },
+      { timeout: 250, interval: 10 }
+    )
+
+    expect(complete).toBe(true)
+    jest.useRealTimers()
```
mpeyper, line 149: This should be in a `catch` block. If the test fails, we don't want to the fake timers to break other tests in the file.

mpeyper:
@chris110408 I'm thinking of merging these tests but keeping #631 open until we get some clarification on what the actual issue is there, as I feel these represent valid ways of using fake timers and they will help ensure we don't accidentally break anything for the current functionality.

I know it's annoying (read my thoughts on it [here](https://github.com/testing-library/react-hooks-testing-library/pull/639#issuecomment-864622639), but would you mind making variations for `native` and `server` renderers?  It'll mostly be copying the file into their test directory and you may need to add a `hydrate` call for the server renderer (see other `server` tests for examples).

src/dom/__tests__/asyncHook.test.ts
chris110408: Copy that makes sense. I will wrap the `jest.useRealTimers()` in finally

mpeyper: lol, yes, I meant `finally`, not `catch`... putting it in `catch` would not be useful at all 🤦

PR #639, opened 2021-06-18
test: Update tests to be compatible with Jest 27

mpeyper:
**What**:

Update test suite to be compatible with `jest` 27.

**Why**:

[The PR to update `kcd-scripts`](https://github.com/testing-library/react-hooks-testing-library/pull/634) is failing because of some changes in `jest` 27 that prevent `beforeEach` and `afterEach` hooks being added after the tests have started to run.  In our test suite, we were doing this in order to test the auto cleanup and error suppression features.  

**How**:

Most of the changes here involve moving the setup and `require` statements outside of the `beforeEach`/`beforeAll` hooks they were bing called it (i.e. after the tests were running) into the `describe` blocks.  While researching how to deal with this, I discovered that `jest` runs each test file in it's own sandbox, so much of the protection I thought using the test hooks were giving us is there within the file which was sufficient for most of the affected tests.

I also took the opportunity to review how the imports were occurring and found a few and were using `require` when a regular `import` would work for that case.  I also found a few unused imports to clean up.

**Checklist**:

- [x] Tests
- [x] Ready to be merged
      

I really want to look at removing the duplication in the test suite.

### DIFF
--- a/src/__tests__/defaultRenderer.pure.test.ts
+++ b/src/__tests__/defaultRenderer.pure.test.ts
@@ -1,0 +1,41 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+import { ReactHooksRenderer } from '../types/react'
+
+describe('default renderer', () => {
+  beforeEach(() => {
+    jest.resetModules()
+  })
+
+  test('should resolve native renderer as default renderer', () => {
+    const expectedRenderer = require('../native/pure') as ReactHooksRenderer
+    const actualRenderer = require('../pure') as ReactHooksRenderer
+
+    expect(actualRenderer).toEqual(expectedRenderer)
+  })
+
+  test('should resolve dom renderer as default renderer', () => {
+    jest.doMock('react-test-renderer', () => {
+      throw new Error('missing dependency')
+    })
+
+    const expectedRenderer = require('../dom/pure') as ReactHooksRenderer
+    const actualRenderer = require('../pure') as ReactHooksRenderer
+
+    expect(actualRenderer).toEqual(expectedRenderer)
+  })
+
+  test('should throw error if a default renderer cannot be resolved', () => {
+    jest.doMock('react-test-renderer', () => {
+      throw new Error('missing dependency')
+    })
+
+    jest.doMock('react-dom', () => {
+      throw new Error('missing dependency')
+    })
+
+    const expectedMessage =
+      "Could not auto-detect a React renderer. Are you sure you've installed one of the following\n  - react-dom\n  - react-test-renderer\nIf you are using a bundler, please update your imports to use a specific renderer.\nFor instructions see: https://react-hooks-testing-library.com/installation#being-specific"
+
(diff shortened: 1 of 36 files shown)


src/__tests__/defaultRenderer.pure.test.ts
```diff
@@ -0,0 +1,41 @@
...
+import { ReactHooksRenderer } from '../types/react'
+
+describe('default renderer', () => {
+  beforeEach(() => {
+    jest.resetModules()
+  })
+
+  test('should resolve native renderer as default renderer', () => {
+    const expectedRenderer = require('../native/pure') as ReactHooksRenderer
+    const actualRenderer = require('../pure') as ReactHooksRenderer
```
mpeyper, line 11: this was the only test where I felt the `require`s in the tests were still appropriate.  I worked around the issue by changing to use the `pure` imports instead.   This left the non-`pure` import uncovered, which is why I introduced the [simplified test file](https://github.com/testing-library/react-hooks-testing-library/pull/639/files#diff-6ad1e6bacfc4e8e27309955adc84b9596acd1c31bdbb575dd8cefbf6ced6a454) for it.

joshuaellis (approved):
Looks like a fairly straight forward refactor.

RE. Duplication, I guess this is where it would be useful to have a set of tests that are called with each renderer type. Could we do `forEach` loops with the specific functions from the library to avoid duplication? Although not exactly "pretty", it could work.

I suppose in terms of structure we'd then hoist tests out of the target and into a specific `test` folder.

mpeyper:
> RE. Duplication, I guess this is where it would be useful to have a set of tests that are called with each renderer type. Could we do forEach loops with the specific functions from the library to avoid duplication? Although not exactly "pretty", it could work.

Yeah, there's also [`jest-in-case`](https://github.com/atlassian/jest-in-case) or the built in [`test.each`](https://jestjs.io/docs/api#testeachtablename-fn-timeout) that could be used instead of a `forEach` loop.  In any case (pun intended), this is likely what we'll end up doing.

In my ideal world, the test _file_ would execute multiple times with different renderers so that the import of on doesn't affect the others (e.g. the cleanup and error suppression with be set up and torn down multiple times), but I've been looking into that it is seems to either be very difficult right now. [`jest-runner-groups`](https://www.npmjs.com/package/jest-runner-groups) seems about the closest, but doesn't expose any ability to know which group is running to switch our the renderer with.

> I suppose in terms of structure we'd then hoist tests out of the target and into a specific test folder.

I was thinking we could just put them in `src/__tests__` with the default renderer tests?  I don't mind having a seperate `test` root for them.  We actually had that before we duplicated everything for the multiple renderers, so bringing it back when we merge the tests again makes sense.

joshuaellis:
> I was thinking we could just put them in src/__tests__ with the default renderer tests? I don't mind having a seperate test root for them. We actually had that before we duplicated everything for the multiple renderers, so bringing it back when we merge the tests again makes sense.

Yeah i think if we're merging them that's the direction  to take

PR #641, opened 2021-06-21
add tests to prove jest useFakeTimers is supported

chris110408:
**What**:
 Add test to show the jest.useFakeTimers() which use with advancing the timers has been supported. However, the reason why `dom-testing-library` implement the feature( jest.useFakeTimers()  work automagically without explicitly advancing the timers) is not clear, a further investigation may continue.

**Why**:
related to #631,

**How**:
we add tests use jest.useFakeTimers()  (wrap with try-catch and finally to make sure the fake timers would not break other tests when test fail,) in dom, native, and server. all test success run 

**Checklist**:

- [x] Documentation updated
- [x] Tests
- [x] Ready to be merged
      

chris110408:
@mpeyper Sorry I had a merge conflict and I am fixing it by  creating this new pull request  😸 
try-catch-finally has been added and the same tests have been used for dom native and server. I think it is ready for merge.
The only thing on those checklist is `Added myself to contributors table`. I am not sure how to do is  Added myself to contributors table.

mpeyper:
I've updated the test to use the new mutil-renderer tests.  I've also moved them into their own file specifically for fake timers so we can expand it further when looking closer at #631.

Thanks again for the effort on this so far, and sorry it took me so long to get back to this.

@all-contributors please add @chris110408 for test

PR #643, opened 2021-06-29
Refactor tests to reduce duplication for different renderers

mpeyper:
**What**:

Refactor tests to reduce duplication for different renderers.

**Why**:

Our test suite has lots of duplication, running the same (or almost the same) test against the `dom`, `native` and `server` renderers.

**How**:

Added a test utility that runs tests against groups of renderers.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

For some reason, the current tests were not having `eslint` warnings appear.  When moving them into `src/__tests__/`, the warnings appeared.  This actually highlighted an issue with the type of the `cleanup` export which is is an `async` function, but wasn't returning a `Promise` in the types.  I've included updating them in this PR so we must remember to merge this as a `fix` rather that the `test` commit.


src/__tests__/utils/runForRenderers.ts
```diff
@@ -0,0 +1,86 @@
...
+global.runForLazyRenderers = function runForLazyRenderers<TRenderer extends Renderer>(
+  renderers: TRenderer[],
+  fn: (getRenderer: () => InferredRenderer<TRenderer>, rendererName: Renderer) => void
+): void {
+  renderers.forEach((renderer) => {
+    describe(renderer, () => {
+      fn(() => rendererResolvers[renderer]() as InferredRenderer<TRenderer>, renderer)
+    })
+  })
+}
```
mpeyper, line 84: This is where most of the magic is happening.

The `runForLazyRenderers` is used in niche situations where we want to control when the `require` is called (i.e. the `autoDetectRenderer` tests.

The `rendererName` argument is used in a handful of cases where the logic is mostly the same and something minor needs to change (generally caused by hydrating the `server` renderer).

src/__tests__/autoDetectRenderer.test.ts
```diff
@@ -0,0 +1,70 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+import { ReactHooksRenderer } from '../types/react'
+
+describe('auto-detect renderer', () => {
```
mpeyper, line 4: This test file replaces the `defaultRenderer` and `defaultRenderer.pure` test files.

src/types/react.ts
```diff
@@ -24,7 +24,7 @@ export type ReactHooksRenderer = {
     options?: RenderHookOptions<TProps>
   ) => RenderHookResult<TProps, TResult>
   act: Act
-  cleanup: () => void
+  cleanup: () => Promise<void>
```
mpeyper, line 27: As mentioned int he PR description, this is fixing the type to match the implementation.

jest.config.js
```diff
@@ -0,0 +1,4 @@
+const { jest: jestConfig } = require('kcd-scripts/config')
+module.exports = Object.assign(jestConfig, {
+  setupFiles: ['<rootDir>/src/__tests__/utils/runForRenderers.ts']
```
mpeyper, line 3: This makes the `global` test utilities available in the tests.

src/__tests__/utils/runForRenderers.ts
```diff
@@ -0,0 +1,86 @@
...
+    ): RenderHookResult<TProps, TResult> {
+      const { hydrate, ...otherUtils } = renderHook<TProps, TResult>(callback, options)
+      hydrate()
+      return {
+        ...otherUtils
+      }
+    },
+    ...otherImports
+  }
+}
```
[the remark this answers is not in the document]
joshuaellis: maybe worth writing this as an inline comment so _others_ know?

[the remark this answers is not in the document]
mpeyper: To clarify, a comment saying why this renderer only exists in the test suite or when you should use it in tests, or both?

I've also been wondering if there is any value in moving this renderer into the `src` and exporting it from the library as well. I'm not sure when it would be used over the `dom` renderer though?

src/__tests__/asyncHook.test.ts
```diff
@@ -0,0 +1,258 @@
+import { useState, useRef, useEffect } from 'react'
+
+describe('async hook tests', () => {
```
mpeyper, line 3: I wouldn't mind having another go at this test file sometime (not in this change).  Currently, it the longest running test by far, clocking almost 20s on my macbook to run all the tests for all the renderers.  We had problems in the past making the delays too small, but that was before the refactoring of the `asyncUtils` which I think has made things more reliable.

src/__tests__/utils/runForRenderers.ts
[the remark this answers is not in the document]
mpeyper: I've added something. Let me know if there's anything else that should be added.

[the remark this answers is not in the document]
joshuaellis: That looks good ⭐

joshuaellis (approved):
It's great to see us loose so much duplication code! Also this way for managing the test suites is also super interesting to see 🚀

PR #658, opened 2021-07-26
Fix #648: Add `setDefaultWaitOptions` function

### LINKED ISSUE #648: Enable setting DEFAULT_INTERVAL and DEFAULT_TIMEOUT for wait* methods
### Describe the feature you'd like:

The wait* methods take optional `interval` and `timeout` values, and fall back to hard coded defaults if not set. 
I would like for the defaults to be configurable.
It's straightforward to wrap the methods, so this is certainly not critical, but I think it would be nice

### Suggested implementation:

Add a `setDefaultWaitOptions` method that can update the options that the `wait*` methods fall back to

### Describe alternatives you've considered:

There could be a way to read from a config file on environment. 
Maybe it would be better to have separate methods for setting them.
If desired, a stack could be implemented to push/pop settings for regions to have different defaults.
I really don't think it needs to be this complicated though

### Teachability, Documentation, Adoption, Migration Strategy:

Users would be able to call the function in a setup file and then use the `wait*` functions normally, but with updated settings

orokanasaru:
**What**:

Fix for #648. Add `setDefaultWaitOptions` function

**Why**:

It simplifies configuration in environments where the defaults may not be optimal

**How**:

* Created a function to override default time options
* Added tests to ensure behavior mimics that of the per-call options
* Re-exported at various levels to match existing patterns and ensure availability regardless of importing at the top level or for a specific renderer

**Checklist**:

- [X] Documentation updated
- [X] Tests
- [X] Ready to be merged
      
Had issues with `docz`. Needed to remove `repo` field from `package.json` and update glob pattern for `files` in `doczrc.js` to build updated documentation. These changes are not included as they presumably do not apply to the pipeline here


mpeyper:
Thanks @orokanasaru for putting this together.  Your comments about docz running on the PR pipeline made me wonder because I thought we had Netlify deploy preview running on All PR branches... Turns out that our Netlify config was never updated when we switched from `master` to `main` for our production branch and we haven't had any doc updates running since May!  So thanks for highlighting that as well.

Unfortunately you're correct about the build issue for docz which appears to be [an open issue](https://github.com/doczjs/docz/issues/1635) for them.  I'm reluctant to release a new feature without the accompanying documentation and I'm more reluctant to remove the repo field from the `package.json` to fix it, so I'm still looking for a way around this to get this in.

Honestly, this is not the first time docz has just randomly broken on us and I've been considering migrating the docs to something else for a while now.  This may be the final trigger for it (we'll see how my investigation tonight into working around it goes).

I'll give the changes a review now anyway.

src/__tests__/asyncHook.test.ts
```diff
@@ -1,6 +1,11 @@
+import { setDefaultWaitOptions } from 'core'
 import { useState, useRef, useEffect } from 'react'
 
 describe('async hook tests', () => {
+  beforeEach(() => {
+    setDefaultWaitOptions({ interval: 50, timeout: 1000 })
+  })
```
mpeyper, line 7: My only hesitation is that there isn't a test that uses the _actual_ default options anymore.  We've used the technique of creating seperate files to test things with global side effects so I'd suggest introducing `asyncHook.customDefaults.ts` (or something like that) with this and the tests you've added and keep this file using the regular old defaults.

src/core/asyncUtils.ts
```diff
@@ -10,8 +10,13 @@ import {
...
 
-const DEFAULT_INTERVAL = 50
-const DEFAULT_TIMEOUT = 1000
+let defaultInterval: number | false = 50
+let defaultTimeout: number | false = 1000
+
+function setDefaultWaitOptions({ interval, timeout }: WaitOptions) {
+  defaultInterval = interval ?? defaultInterval
+  defaultTimeout = timeout ?? defaultTimeout
+}
```
mpeyper, line 19: Do we need some way of resetting the defaults back to the initial values?

I'm just thinking about someone that wants to update the defaults for a single test file so they set it in the `beforeAll` and want to reset it in the `afterAll`.  Because we don't export the initial default values, they would have to just hard code the documented default values and hope we never change them.

The options I see are:
1. export some constants for them to use with `setDefaultWaitOptions`
2. export a `resetDefaultWaitOptions` utility

Or maybe both?  Or maybe there's another option I'm not seeing?

src/core/index.ts
```diff
@@ -82,4 +82,11 @@ function createRenderHook<
...
 }
 
-export { createRenderHook, cleanup, addCleanup, removeCleanup, suppressErrorOutput }
+export {
+  createRenderHook,
+  cleanup,
+  addCleanup,
+  removeCleanup,
+  suppressErrorOutput,
+  setDefaultWaitOptions
```
mpeyper, line 91: I feel like the worst kind of person to nitpick this on a review, but this one is in the opposite order to the others:

```suggestion
  setDefaultWaitOptions,
  suppressErrorOutput
```

I hate that this bothers me 😞

silviuaavram:
I am also interested into a better workaround for that `repo.browsetemplate.replace` issue in docz.

mpeyper:
@silviuaavram sorry, I don't have any good news.  I'm currently porting our docs over to [docusaurus](https://docusaurus.io/) 🤷‍♂️

silviuaavram:
Cool :). While doing this you could probably work around as I did now, https://github.com/doczjs/docz/issues/1635#issuecomment-888953896. Let me know if this works for you!

src/core/asyncUtils.ts
joshuaellis: I think exporting the constants makes more sense than a function imo

mpeyper: I've been wondering about options into `renderHook`? A render helper can then be extracted like normal for shared setup.

mpeyper:
Hi @orokanasaru, just checking in to see where you're at with this.  Is it still something you want to pursue?

src/core/asyncUtils.ts
orokanasaru: @mpeyper, can you expand on your idea? I will export initial constants otherwise

orokanasaru:
@mpeyper, you said you did not want to release until documentation issues you resolved the documentation issues. If that is no longer a blocker, I can return to finish this up

PR #669, opened 2021-08-04
feat: add customRender option

ykan:
**What**:

Add a `customRender` option to custom `TestComponent` of `src/helpers/createTestHarness.tsx`

**Why**:

When write my custom hook, I have to use a dom ref. And it's hard to mock a dom element when I write tests. For example:

```ts
import React from 'react'

import { useForkRef } from './useForkRef'

export function useCustom<T extends HTMLElement>(ref: React.Ref<T>) {
  const imgRef = React.useRef<T>(null)
  const handleRef = useForkRef(ref, imgRef)

  React.useEffect(() => {
    if (!imgRef.current) {
      return
    }
    const rect = imgRef.current.getBoundingClientRect()
  }, [])

  return {
    handleRef,
  }
}
```

**How**:

Add a Custom param to the function `createTestHarness`.

**Checklist**:

- [x] Documentation updated
- [x] Tests
- [x] Ready to be merged
      

mpeyper:
Hi @ykan,

Thanks for taking the time to put this together.  Generally I would recommend raising an issue before putting effort into a PR, but as it's here now, I'm happy to talk about it here and with this implementation in mind.

I'm inclined to suggest that this is an example of when the requirements of the test go passed what this library aims to do, which is to simplify testing custom hooks.  In my opinion, the example test is confusing for me coming in and reading it after the fact.  This is also true for me when reading the documentation explaining the feature, so I'm concerned about the teachability of this feature.

Our goto advice when you need this level of control and composition of hooks, components and refs is usually to just use `@testing-library/react` and write a component in your test file that creates a ref, calls the hook and passes to the DOM element.

Something else worth mentioning is that often a combination of `@testing-library/react-hooks` and `@testing-library/react` can be used to test the returned values from the hook work when rendered with real components, e.g:

```ts
import React, { useRef } from "react";
import { render } from "@testing-library/react";
import { renderHook } from "@testing-library/react-hooks/dom";

function useDomRef() {
  const domRef = useRef<HTMLElement>();
  const refCallback = (el: HTMLElement | null) => {
    if (el) {
      domRef.current = el;
    }
  };

  return { refCallback, domRef };
}

test("render will get a dom ref", () => {
  const { result } = renderHook(() => useDomRef());
  render(<div ref={result.current.refCallback} />);
  
  expect(result.current.domRef.current).toBeDefined();
});
```

This doesn't work for all situations, like the run once effect used in the PR description, but it does cover many use cases where the interaction in the rendered components update the hook state.

In general, I don't think I like the ergonomics of the `customRender` option being proposed here, but as I'm writing this response, I am wondering how common it is the have hooks return refs that need to be applied to a DOM element before the hook functionality actually works.  Perhaps we need to look the intention of the use case and see if there is anything we want to do to help support it.  @joshuaellis, thoughts?

mpeyper:
Also, your tests are failing because you've left the default value for the `suspense` parameter of `createTestHarness` which is now never falling back to the default because it's always being passed in.  If you drop the `= true` you should pass the coverage check again.

joshuaellis:
>In general, I don't think I like the ergonomics of the customRender option being proposed here, but as I'm writing this response, I am wondering how common it is the have hooks return refs that need to be applied to a DOM element before the hook functionality actually works. Perhaps we need to look the intention of the use case and see if there is anything we want to do to help support it. @joshuaellis, thoughts?

I think its somewhat common to be returning `refs` to do xyz especially if you're working with animation, webgl etc. However, I do think if you're using hooks that return refs, the hook normally manipulates the element attached to the ref thus it's just as easy to test against this with `@testing-library/react`.

On top of that, afaik you can only use refs when using `ReactDOM` as the test renderer, `react-test-renderer` doesn't actually handle refs as nothing is physically mounted / created so you have to pass a second arg to the test-renderer, see [this](https://reactjs.org/docs/test-renderer.html#ideas) for more information. This will probably cause a bit of confusion if it doesn't work the same way across the library depending on which renderer you use, and we have auto select so some people might not even think twice about it...

So I do feel the same as @mpeyper in the sense, I think this goes beyond the scope of the library.

ykan:
Forgive my poor English. 
In my case, it's hard to test my code in `React.useEffect`. Actually, in @mpeyper 's code using `@testing-library/react` will not work when `React.useEffect(() => {}, [])` is working like `componentDidMount`.  Just like the code here [Example hook tests](https://github.com/ykan/robot-img/blob/8bded519331e8ca55070a134af982c54435ae88a/packages/react-img/src/useImg.test.tsx#L75), I have to invode `rerender` to test my code in the useEffect function .
Of course, it can write more code using `@testing-library/react`, it just a litte inconvenient.

PR #671, opened 2021-08-07
Pr/fix typo in docs

masious:
**What**:

The change is a typo fix.

**Why**:

Currently, the documentation is implying that the return value of the `renderHook` result is the same as `current` which is wrong. This PR fixes the implication.

**How**:

Just by changing a character in the docs!

**Checklist**:

- [x ] Documentation updated
- [x ] Ready to be merged
- [x ] Added me to the contributors' table

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -584,6 +584,15 @@
       "contributions": [
         "test"
       ]
+    },
+    {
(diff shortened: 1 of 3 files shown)


joshuaellis (approved):
Thanks for the contribution! ⭐

PR #682, opened 2021-08-30
fix(async-utils): prevent timeout and interval checks in wait from leaving open handles

mpeyper:
**What**:

This issue was raised in [discord](https://discord.com/channels/723559267868737556/723852521734668299/880288543501586492) but the use of `Promise.race` has also been flagged as potentially an issue with fake timers in #631.

(I'm going to use "timeout" a lot in here that had slightly different meanings so I hope it doesn't get too confusing).

The timeout that get created to stop waiting after the async utils timeout is exceeded and the timeout that created for running intverval checks when the hook doesn't rerender fast enough are not being cleared when the wait condition is being satisfied and the promise resolves before they fire.  This is leaving open handles which `jest` is warning about.

Interestingly, the warning only appears when the `testEnvironment` is not set to `jsdom` in the `jest` config (like in our test suite), which I haven't been able to properly identify why, but I'm assuming `jsdom` must be intercepting these handle somehow and cleaning them up, which doesn't occur when using the `node` test environment.

**Why**:

We want to be good citizens and clean up after outselves.

**How**:

The `Promise.race` was not the actual cause, but the use of racing promises and letting the losing promise just run to completion was not helping here.  While the implementation allowed for each part to remain relatively simple, it made it difficult to cancel the loser when the winner resolved.

The new implementation introduces a `timeoutSignal` that controls the timeouts and wraps the promise in a way that is aware of when it completely as clears the pending timeouts so they are no longer left open.   This means there is a little bit more ceremony in using it and we need to check the whether it timed out in a few more places to ensure things exit cleanly.

**Checklist**:

- [ ] Documentation updated
   - N/A
- [ ] Tests
  - The existing tests are passing without modification, but I'm unsure if we can/want to run specific tests in a non-`jsdom` test environment with `--detectOpenHandles` to ensure it stays working (I don't particularly want to introduce any more variations of the test suite)
- [x] Ready to be merged
      

joshuaellis (approved):

joshuaellis:
I think we're okay without the extra tests, I agree they'll add more complexity to our test suite, I think if we begin seeing this as a pain point, we can then address adding specific tests.

mpeyper:
I just made a small change to rename `createTimeoutSignal` to `createTimeoutController`.  To be honest, I'm not sure what to call it, but it felt like it was behaving more like the `AbortController` itself that the `AbortController.signal` that I originally named it after.

PR #688, opened 2021-09-06
Fix #631, add  jestFakeTimersAreEnabled and use it to detect faketimer in createTimeoutController

### LINKED ISSUE #631: `waitFor` doesn't work if jest fake timers are used

chris110408:
**What**:

Fixes #631, add jestFakeTimersAreEnabled and use it detect the useFakeTimer.
 when useFakeTimer has been detected and jest.advanceTimersByTime will advance time to the timeout-1
test:'should waitFor arbitrary expectation to pass when fake timers are not advanced explicitly' has been added

**Why**:
we need to add jestFakeTimersAreEnabled function the same function used on `@testing-library/dom`
and use it to detect the fake timer and use advanced time  to make sure time is not stuck

**How**:
The problem was when the fake timer is used. the time will not move forward unless using the advanced time. 
#682  `createTimeoutController` has been used to replace `Promise.race`. 
the problem is to detect the fake timber and use advance time to move time forward to the last ms until the timeout. 

**Checklist**:

- [x] Documentation updated
- [x] Tests
- [x] Ready to be merged
      
- [ ] Added me to contributors table.  @mpeyper my status is test contributor only could you please update my status to  the code and test


src/helpers/createTimeoutController.ts
```diff
@@ -19,6 +20,10 @@ function createTimeoutController(timeout: WaitOptions['timeout']) {
             timeoutCallbacks.forEach((callback) => callback())
             resolve()
           }, timeout)
+
+          if (jestFakeTimersAreEnabled()) {
+            jest.advanceTimersByTime(timeout - 1)
```
mpeyper, line 25: I'm not sure if this implementation will work in all cases.  Consider if a condition is set to resolve just after the configured timeout.  For example overall timeout of 100ms and 50ms interval checks and check passes at 101ms:

```ts
    test('should reject if timeout is passed close to when promise resolves', async () => {
      const { waitFor } = renderHook(() => null)

      let actual = 0
      const expected = 1

      setTimeout(() => {
        actual = expected
      }, 101)

      let complete = false

      await expect(
        waitFor(
          () => {
            expect(actual).toBe(expected)
            complete = true
          },
          { timeout: 100, interval: 50 }
        )
      ).rejects.toThrow(Error('Timed out in waitFor after 100ms.'))

      expect(complete).toBe(false)
    })
```

chris110408: I see, in this case the sudo code will be 

```js
  if(jestFakeTimersAreEnabled){
  while(intervalSignal not finish){
      jest.advanceTimersByTime(interval)
      if(totalTime > Timeout){
           timeoutController.timedOut = true
      }
  }
}

```

But I am not sure how to implement it now

chris110408: The problem has fixed and test has been added

mpeyper: I've been playing around with this a bit I think part of the reason it's an issue is that both the `interval` and the `timeout` controllers are advancing time, so for a 100ms `timeout` and a 50ms `interval` there is a total of 197ms (`(100-1) + 2 * (50-1)`) advanced (might be 3x the interval, because it's not quite enough to timeout with just 2).  The other issue is the way the promises work vs the timeouts where the `.then` fires after the timeout , regardless of whether it would have happened first with a real timer (I think because it's in the microtask queue but the timeout fires synchronously when the timers are advanced, but it's getting deeper than my JS knowledge goes, so I'm guessing).

I've managed to get something that can pass all the existing async tests, except one, but with a fake timers:

```
        if (allowFakeTimers && jestFakeTimersAreEnabled()) {
          const advanceTime = async (currentMs: number) => {
            if (currentMs < timeout) {
              jest.advanceTimersByTime(1)
  
              await Promise.resolve()
  
              if (finished) {
                return
              }
              await advanceTime(currentMs + 1)
            }
          }

          advanceTime(0)
        }
```

`allowFakeTimers` gets passed into `createTimoutController` and is `true` for the `interval` controller and `typeof interval !== 'number'` for the `timeout` controller (so that time still advances when interval checks are disabled, e.g. `waitForNextUpdate`).  `finished` is a `boolean` that gets set before the `timeoutController.cancel()` in the `.finally` on the end of the promise chain.

The one test that fails is failing because it has both `interval` and `timeout` disabled so there are no timeout controllers to actually advance timers, but the test is still trying to rely on `waitForNextUpdate` rather than advancing the timer itself, so part of me feels like it's an invalid scenario anyway.

mpeyper: Note: I had typed this out in a stale browser and haven't looked at the more recent changes.  I'll leave the comment for you to digest, but I'll also take a look later and it might not be relevant anymore.

chris110408: That is a great Idea

The PR has been updated.

I revised the advanceTime to 
```js
const advanceTime = async () => {
    jest.advanceTimersByTime(timeout as number)
    await Promise.resolve()
  }
```

I have a question why do we need to use `await Promise.resolve()` here

and I also move the `DEFAULT_TIMEOUT` to createTimeoutController file

src/helpers/createTimeoutController.ts
```diff
@@ -1,6 +1,11 @@
 import { WaitOptions } from '../types'
+import { jestFakeTimersAreEnabled } from './jestFakeTimersAreEnabled'
+const DEFAULT_TIMEOUT = 1000
 
-function createTimeoutController(timeout: WaitOptions['timeout']) {
+function createTimeoutController(
+  timeout: WaitOptions['timeout'] = DEFAULT_TIMEOUT,
```
chris110408, line 6: @mpeyper This line is little confusing. Since some time it is not timeout but interval 
We may need to reword it and give WaitOptions['timeout'] |WaitOptions['interval'] type for it.

mpeyper: It's probably unlikely the types for `timeout` and `interval` will ever diverge, but I see your point.

Perhaps were better off typing it explicitly as what it allows (i.e. `number | false`) and because this function is actually not related to `WaitOptions` other than by coincidence.

chris110408: I changed the type to number | false. I think it is ready to merge now

mpeyper:
@chris110408 I've just pushed some more tests for the fake timers test suite ([they are the same as the regular async utils, except one](https://github.com/testing-library/react-hooks-testing-library/pull/688#discussion_r704083820)).

Theoretically all these tests should pass, when fake timers are enabled.  With your current changes, most are timing out (the CI agents are not handling this very well as it's taking a long time for all of the tests to time out).

src/__tests__/asyncHook.fakeTimers.test.ts
```diff
@@ -51,8 +74,273 @@ describe('async hook (fake timers) tests', () => {
...
+    // eslint-disable-next-line jest/no-disabled-tests
+    test.skip('should not reject when waiting for next update if timeout has been disabled', async () => {
+      const { result, waitForNextUpdate } = renderHook(() => useSequence(['first', 'second'], 1100))
+
+      expect(result.current).toBe('first')
+
+      await waitForNextUpdate({ timeout: false })
+
+      expect(result.current).toBe('second')
+    })
```
mpeyper, line 158: this is the one skipped test.  `waitForNextUpdate` has `interval` implicitly set to `false` as well as `timeout`, so without anything advancing timers it' just hangs.  I'm not sure how to deal with this case or if we even should yet.

src/core/asyncUtils.ts
```diff
@@ -20,11 +20,11 @@ function asyncUtils(act: Act, addResolver: (callback: () => void) => void): Asyn
       return callbackResult ?? callbackResult === undefined
     }
 
-    const timeoutSignal = createTimeoutController(timeout)
+    const timeoutSignal = createTimeoutController(timeout as number | boolean, false)
```
mpeyper, line 23: If `interval` is `false` (like it is implicitly for `waitForNextUpdate`, but can also be set by the user for the other utils) you really want to be passing `true` here.  For my local testing, I've been using

```suggestion
    const timeoutSignal = createTimeoutController(timeout as number | boolean, !interval)
```

This way, either `intervalSignal` or `timeoutSignal` will advance the timers, but not both.

src/core/asyncUtils.ts
```diff
@@ -10,8 +10,8 @@ import {
 import { createTimeoutController } from '../helpers/createTimeoutController'
 import { TimeoutError } from '../helpers/error'
 
-const DEFAULT_INTERVAL = 50
 const DEFAULT_TIMEOUT = 1000
+const DEFAULT_INTERVAL = 50
 
 function asyncUtils(act: Act, addResolver: (callback: () => void) => void): AsyncUtils {
   const wait = async (callback: () => boolean | void, { interval, timeout }: WaitOptions) => {
```
mpeyper, line 17: using `WaitOptions` is a bit problematic here as the `undefined` has actually been resolved by this point for each of the actual util wrappers.

I've been playing with introducing a `ResolvedWaitOptions` type here to remove it so we don't have to do any casting below:
```suggestion
type ResolvedWaitOptions = { [T in keyof WaitOptions]-?: WaitOptions[T] }

function asyncUtils(act: Act, addResolver: (callback: () => void) => void): AsyncUtils {
  const wait = async (callback: () => boolean | void, { interval, timeout }: ResolvedWaitOptions) => {
```

There's probably a utility type somewhere (built in or `type-fest` that would give a nice generic wrapper for this.  This was just playing around to see if it was possible.

src/helpers/createTimeoutController.ts
```diff
@@ -1,6 +1,6 @@
-import { WaitOptions } from '../types'
+import { jestFakeTimersAreEnabled } from './jestFakeTimersAreEnabled'
 
-function createTimeoutController(timeout: WaitOptions['timeout']) {
+function createTimeoutController(timeout: number | boolean, allowFakeTimers: boolean) {
```
mpeyper, line 3: `true` is not a valid value, only `number | false`

```suggestion
function createTimeoutController(timeout: number | false, allowFakeTimers: boolean) {
```

This would mean you would not need to cast it as a number below.

src/core/asyncUtils.ts
mpeyper: Yep, you can totally just replace `WaitOptions` with `Required<WaitOptions>` and remove the casts

```suggestion
function asyncUtils(act: Act, addResolver: (callback: () => void) => void): AsyncUtils {
  const wait = async (callback: () => boolean | void, { interval, timeout }: Required<WaitOptions>) => {
```

src/core/asyncUtils.ts
```diff
@@ -20,11 +20,11 @@ function asyncUtils(act: Act, addResolver: (callback: () => void) => void): Asyn
       return callbackResult ?? callbackResult === undefined
     }
 
-    const timeoutSignal = createTimeoutController(timeout)
+    const timeoutSignal = createTimeoutController(timeout as number | boolean, false)
 
     const waitForResult = async () => {
       while (true) {
-        const intervalSignal = createTimeoutController(interval)
+        const intervalSignal = createTimeoutController(interval as number | boolean, true)
```
mpeyper, line 27: just as a suggestion for readability, I prefer boolean params to be named options as it makes it clearer what they do when reading from this side:

```ts
const intervalSignal = createTimeoutController(interval, { allowFakeTimers: true })
```

src/__tests__/asyncHook.fakeTimers.test.ts
mpeyper: I think I have just found a way to get this case to pass as well.

src/core/asyncUtils.ts
```diff
@@ -20,11 +20,11 @@ function asyncUtils(act: Act, addResolver: (callback: () => void) => void): Asyn
...
         timeoutSignal.onTimeout(() => intervalSignal.cancel())
 
         await intervalSignal.wrap(new Promise<void>(addResolver))
```
mpeyper, line 30: I've just noticed that I didn't rename these to `timeoutController` and `intervalController` when I renamed `createTimeoutSignal` to `createTimeoutController` 😅

mpeyper:
@chris110408, you probably got a notification about it (but in case anyone is reading along) but [I've pushed up a branch](https://github.com/chris110408/react-hooks-testing-library/pull/1) with most of the suggested changes here and a modified advance timers implementation that passes all the tests (at least locally for me).

chris110408:
> @chris110408, you probably got a notification about it (but in case anyone is reading along) but [I've pushed up a branch](https://github.com/chris110408/react-hooks-testing-library/pull/1) with most of the suggested changes here and a modified advance timers implementation that passes all the tests (at least locally for me).

Thank you. I will look at those change today

src/helpers/createTimeoutController.ts
chris110408: 👍

src/core/asyncUtils.ts
chris110408: Yeah, 'intervalController' and 'timeoutController' are a better names , I will change them
👍

src/core/asyncUtils.ts
chris110408: 👍

src/core/asyncUtils.ts
chris110408: 👍

src/core/asyncUtils.ts
chris110408: 👍

src/__tests__/asyncHook.fakeTimers.test.ts
```diff
@@ -51,8 +74,272 @@ describe('async hook (fake timers) tests', () => {
...
+      const { result, waitForNextUpdate } = renderHook(() => useSequence(['first', 'second']))
+
+      expect(result.current).toBe('first')
+
+      await expect(waitForNextUpdate({ timeout: 10 })).rejects.toThrow(
+        Error('Timed out in waitForNextUpdate after 10ms.')
+      )
+    })
+
+    test('should not reject when waiting for next update if timeout has been disabled', async () => {
```
chris110408, line 149: @mpeyper  The skip has been removed and I made some other revise to make this test pass

src/helpers/createTimeoutController.ts
```diff
@@ -12,22 +30,28 @@ function createTimeoutController(timeout: WaitOptions['timeout']) {
...
         timeoutController.onTimeout(resolve)
-
         if (timeout) {
           timeoutId = setTimeout(() => {
             timeoutController.timedOut = true
             timeoutCallbacks.forEach((callback) => callback())
             resolve()
           }, timeout)
         }
+        if (jestFakeTimersAreEnabled() && allowFakeTimers) {
```
chris110408, line 40: I have to move  this one out of the timeout, otherwise the time will not move forward
for test 'should not reject when waiting for next update if timeout has been disabled'

src/helpers/createTimeoutController.ts
```diff
@@ -1,8 +1,27 @@
...
+function createTimeoutController(timeout: number | false, options: { allowFakeTimers: boolean }) {
   let timeoutId: NodeJS.Timeout
   const timeoutCallbacks: Array<() => void> = []
+  let finished = false
+
+  const { allowFakeTimers = false } = options
+
+  const advanceTime = async (currentMs: number) => {
+    // eslint-disable-next-line no-negated-condition
+    if (currentMs < (!timeout ? DEFAULT_TIMEOUT : timeout)) {
```
chris110408, line 14: when timeout is false we have to give it a DEFAULT_TIMEOUT
for test 'should not reject when waiting for next update if timeout has been disabled'

chris110408:
it passed all validation on my local and do not know why it do not pass checker here
image

mpeyper:
I'll take a look again over the weekend.  That looks like the same error I saw when the tests were just running really slowly before timing out, but maybe there is something else going wrong on the build (although I'm not sure what it could be).

chris110408:
> I'll take a look again over the weekend. That looks like the same error I saw when the tests were just running really slowly before timing out, but maybe there is something else going wrong on the build (although I'm not sure what it could be).

chris110408:
> I'll take a look again over the weekend. That looks like the same error I saw when the tests were just running really slowly before timing out, but maybe there is something else going wrong on the build (although I'm not sure what it could be).

I checked everything on my local everything looks fine

src/__tests__/asyncHook.fakeTimers.test.ts
```diff
@@ -51,8 +74,272 @@ describe('async hook (fake timers) tests', () => {
...
+    test('should reject if timeout exceeded when waiting for next update', async () => {
+      const { result, waitForNextUpdate } = renderHook(() => useSequence(['first', 'second']))
+
+      expect(result.current).toBe('first')
+
+      await expect(waitForNextUpdate({ timeout: 10 })).rejects.toThrow(
+        Error('Timed out in waitForNextUpdate after 10ms.')
+      )
+    })
+    //  the test  run successfully after remove skip in local. However, it will not pass the checkers on github
```
chris110408, line 148: @mpeyper The checker fail because of this test. It works in my local. after I added the skip, All checker passed.

mpeyper:
I'm just reading a few things ([here](https://stackoverflow.com/questions/69118749/github-action-error-while-running-unit-test-pipeline), [here](https://github.com/nodejs/node/issues/40030) and [here](https://bugs.chromium.org/p/v8/issues/detail?id=12188) that suggest te build error we were getting was actually a node error (from v8) and just very unlucky timing on our part.

I'm going to try reenabling that test a see if it passes now, but will revert if it's still an issue.

chris110408:
> I'm just reading a few things ([here](https://stackoverflow.com/questions/69118749/github-action-error-while-running-unit-test-pipeline), [here](https://github.com/nodejs/node/issues/40030) and [here](https://bugs.chromium.org/p/v8/issues/detail?id=12188)) that suggest te build error we were getting was actually a node error (from v8) and just very unlucky timing on our part.
> 
> I'm going to try reenabling that test a see if it passes now, but will revert if it's still an issue.
> 
> **Update:** it worked!

NIce.

mpeyper:
> > I'm just reading a few things ([here](https://stackoverflow.com/questions/69118749/github-action-error-while-running-unit-test-pipeline), [here](https://github.com/nodejs/node/issues/40030) and [here](https://bugs.chromium.org/p/v8/issues/detail?id=12188)) that suggest te build error we were getting was actually a node error (from v8) and just very unlucky timing on our part.
> > I'm going to try reenabling that test a see if it passes now, but will revert if it's still an issue.
> > **Update:** it worked!
> 
> Nice. Is there any thing I need to double check before we merge this PR?

I need to properly review it (spent most of my time looking at the build issue the other night), but I'm thinking of putting this out in a beta release first.

mpeyper:
Thanks a lot @chris110408.  I've made a few, most cosmetic changes.

The biggest change was no using `DEFAULT_TIMEOUT` for advancing timers when `timeout = false` to rather keep advancing while the promise is outstanding.

The only other thing I noticed that might have been an issue (no tests were failing) was that timers were being advance in both the timeout and interval controllers since they were moved out of the the `if (timeout)` check for the `timeout = false, interval = false` test case, but that just meant we needed to pick one to run the timers.

I'm going to push out an alpha build for this now.

mpeyper (approved):

mpeyper:
Just wondering if we need some a note in the [async utils docs](https://react-hooks-testing-library.com/reference/api#async-utilities) about fake timers?

mpeyper:
I've got a fairly complicated real world example that I thought would be fixed by this change, but after testing out the alpha it's still not working.  I'll try to get a boiled down version onto a test to help work through it, but so far it appears as though the microtask queue is not flushing as we advance timers.  I'm not sure why this is occurring for the setup yet, but wanted to drop a comment here in case anyone else tries it with a more real world example and is also running into issues.

chris110408:
> Thanks a lot @chris110408. I've made a few, most cosmetic changes.
> 
> The biggest change was no using `DEFAULT_TIMEOUT` for advancing timers when `timeout = false` to rather keep advancing while the promise is outstanding.
> 
> The only other thing I noticed that might have been an issue (no tests were failing) was that timers were being advance in both the timeout and interval controllers since they were moved out of the the `if (timeout)` check for the `timeout = false, interval = false` test case, but that just meant we needed to pick one to run the timers.
> 
> I'm going to push out an alpha build for this now.

@mpeyper Could you please add me to the code contributor list?

chris110408:
> Just wondering if we need some a note in the [async utils docs](https://react-hooks-testing-library.com/reference/api#async-utilities) about fake timers?

I could work on that If we need to

chris110408:
> I've got a fairly complicated real world example that I thought would be fixed by this change, but after testing out the alpha it's still not working. I'll try to get a boiled down version onto a test to help work through it, but so far it appears as though the microtask queue is not flushing as we advance timers. I'm not sure why this is occurring for the setup yet, but wanted to drop a comment here in case anyone else tries it with a more real world example and is also running into issues.

will look at it this weekend

mpeyper:
> > Thanks a lot @chris110408. I've made a few, most cosmetic changes.
> > The biggest change was no using `DEFAULT_TIMEOUT` for advancing timers when `timeout = false` to rather keep advancing while the promise is outstanding.
> > The only other thing I noticed that might have been an issue (no tests were failing) was that timers were being advance in both the timeout and interval controllers since they were moved out of the the `if (timeout)` check for the `timeout = false, interval = false` test case, but that just meant we needed to pick one to run the timers.
> > I'm going to push out an alpha build for this now.
> 
> @mpeyper Could you please add me to the code contributor list?

[I did](https://github.com/chris110408/react-hooks-testing-library/blob/d62357269ae6b0e966467b5eccba2a605e59afdb/.all-contributorsrc#L585), it's just in the alpha branch for now.

> I could work on that If we need to

Thanks. Just target the alpha branch if it's not in main when you make the PR.

> will look at it this weekend

Hopefully I will have booked down the example by then. I'll post it here if I do. I'm actively working on this too as the hook/test in question is part of my day job.

chris110408:
> > @mpeyper Could you please add me to the code contributor list?
> 
> [I did](https://github.com/chris110408/react-hooks-testing-library/blob/d62357269ae6b0e966467b5eccba2a605e59afdb/.all-contributorsrc#L585), it's just in the alpha branch for now.
> 
> > I could work on that If we need to
> 
> Thanks. Just target the alpha branch if it's not in main when you make the PR.
> 
> > will look at it this weekend
> 
> Hopefully I will have boiled down the example by then. I'll post it here if I do. I'm actively working on this too as the hook/test in question is part of my day job.
Thank you

mpeyper:
> I've got a fairly complicated real world example that I thought would be fixed by this change, but after testing out the alpha it's still not working. I'll try to get a boiled down version onto a test to help work through it, but so far it appears as though the microtask queue is not flushing as we advance timers. I'm not sure why this is occurring for the setup yet, but wanted to drop a comment here in case anyone else tries it with a more real world example and is also running into issues.

So after a few hows of replacing portions of this very complex hook setup, I've managed to narrow it down to an API call being made with `node-fetch` that is being mocked with `msw`.  If I replace that API call with a promise that resolves after a short timeout, the test works as expected.

I'm out of steam today to look at this anymore, but I'll try to setup  a small example just with `node-fetch` and `msw` and see I can replicate it without everything else going on in this hook (it's also got event emitters, intervals, memoized functions and redux dispatching all thrown in mix) sometime over the next few days.

icatalina:
Hi! thanks for fixing this! We are facing this issue, when are we planning to release it? thanks 🙇

icatalina:
After a bit of testing with the alpha version, it doesn't seem to work.

```tsx
waitFor(() => {
  console.log('check');

  expect(true).toBe(false);
}, { timeout: 30_000, interval: 1_000 });
```

We can see the `check` printing all immediately, we would expect it print once every second for 30s.

mpeyper:
Hi @icatalina, I'm not sure this will ever see a full release at this point, given [what is happening over in RTL](https://github.com/testing-library/react-testing-library/pull/991). I believe `renderHook` is available now ["in a stable release"](https://github.com/testing-library/react-testing-library/pull/991#issuecomment-981085443), so you should probably check that out.  As a bonus, their version of `waitFor` already supports fake timers from jest.

icatalina:
Ohhh!! I had no idea! thanks heaps for pointing that out! :)

PR #706, opened 2021-09-29
docs: update readme with the more common API

kentcdodds:
**What**:

Updates the README example

**Why**:

Had someone in one of my workshops say it was confusing that the example in this README was different from the more regular way to test hooks.

**How**:

updated README

**Checklist**:

- [x] Documentation updated
- [ ] Tests N/A
- [x] Ready to be merged
      
- [ ] Added myself to contributors table N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ import { renderHook, act } from '@testing-library/react-hooks'
 import useCounter from './useCounter'
 
 test('should increment counter', () => {
-  const { result } = renderHook(() => useCounter())
+  const { result } = renderHook(useCounter)
 
   act(() => {
     result.current.increment()


mpeyper:
Hey @kentcdodds,

This example was actually intentional as simply passing the function, while technically possible in this case, goes against two of my goals for this library:

1. make testing hooks resemble using hooks as closely as possible
2. be accessible to beginners

To point 1, when calling hook functions in components, we can't pass a function reference, we must call it.  In my opinion, `const { result } = renderHook(() => useCounter())` more closely matches how I would use this hook in a component than `const { result } = renderHook(useCounter)` does.

To point 2, this is s very simple hook and could work if the callback is removed, however, experienced JS devs may be comfortable doing this, but I know many juniors that struggle with the idea of passing around function references (yes, I'm aware the callback _is_ a function reference, but to a beginner it can often be seen as just syntax for some things).  The flip side here is that someone that is more comfortable with passing function references should be able to identify the possibility here and make that adjustment to their tests if they want to.

Also as soon as you need to pass an argument to the function, it needs to be expanded again:

```diff
- const { result } = renderHook(useCounter)
+ const { result } = renderHook(() => useCounter(2))
```

vs

```diff
+ const { result } = renderHook(() => useCounter())
+ const { result } = renderHook(() => useCounter(2))
```

The second example is a much easier change to understand and make for someone learning than without needing to google "how to pass arguments to renderHook?"

> the more regular way to test hooks.

It's interesting that this is considered the more regular way as I've always thought of this simplified version as the exception to the rule:

```ts
const { result } = renderHook(useCounter) // no args

const { result } = renderHook(() => useCounter(2)) // static args

const { result } = renderHook(({ step }) => useCounter(step), { initialProps: { step: 3 } }) // one option for dynamic args

const step = 4
const { result } = renderHook(() => useCounter(step)) // another option for dynamic args

  const [count, increment] = useCounter()
  return { count, increment } // common technique for making tuple results more readable
})

  const counter = useCounter()
  return useCountdownTimer(counter) // common technique for composing multiple hooks in an integration test
})
```

Of course there is a whole bunch of my opinion wrapped up in here, and I won't pretend you don't waaaaaay more experience with learners than me, so happy to consider this if you still feel the change is significant improvement with the reasoning above.

kentcdodds:
Fair points 👍 We can leave it as-is :) Maybe I'll change my exercise a bit to reduce the confusion on my end! Thanks :)

PR #747, opened 2021-12-06
Update Act type signature

andersonvom:
**What**:

Update `Act` typescript signatures to be more encompassing.

**Why**:

`Promise<void | undefined>` is too restricted for the `callback`
argument taken by `act()` as certain libraries expose hooks that have
async functions that do return values other than `undefined` or `void`.

For example, `react-query` exposes a `fetchNextPage()` async function
that returns a `Promise<SomeLongResult<T, E>>`. The following snippet
illustrates the issue of the current signature of `act()`:

```ts
// after rendering the hook
const hook = renderHook(() => useCustomHook());

// we're forced to write this
await act(() => hook.result.current.fetchNextPage().then(() => {}));

// instead of one of the simpler ways
await act(hook.result.current.fetchNextPage);
await act(() => hook.result.current.fetchNextPage());
```

**How**:

This changes the current signature to accept a function that returns a
`Promise<any>`. It also changes the sync callback signature to `any`
just to keep both signatures consistent and avoid any confusion as to
why the types are different.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -602,6 +602,15 @@
       "contributions": [
         "doc"
       ]
+    },
+    {
+      "login": "andersonvom",
+      "name": "Anderson Mesquita",
+      "avatar_url": "https://avatars.githubusercontent.com/u/69922?v=4",
+      "profile": "https://github.com/andersonvom",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "skipCi": true,
--- a/README.md
+++ b/README.md
@@ -146,9 +146,9 @@ to test against. It also does not come installed with a specific renderer, we cu
 [`react-test-renderer`](https://www.npmjs.com/package/react-test-renderer) and
 [`react-dom`](https://www.npmjs.com/package/react-dom). You only need to install one of them,
 however, if you do have both installed, we will use `react-test-renderer` as the default. For more
-information see the [installation docs](https://react-hooks-testing-library.com/installation#renderer).
-Generally, the installed versions for `react` and the selected renderer should have matching
-versions:
+information see the
+[installation docs](https://react-hooks-testing-library.com/installation#renderer). Generally, the
+installed versions for `react` and the selected renderer should have matching versions:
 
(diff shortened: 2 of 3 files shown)


joshuaellis:
This typing if I remember correctly, comes straight from @types/react. 

So personally, I'm not sure we should be changing it for our library because that would be confusing for others coming from similar libraries like @testing-libray/react...

mpeyper:
Hi @andersonvom,

Thanks for taking the time to look into this and put this together.  Unfortunately I won't be accepting it today as the act interface is designed to be compatible with both [react-dom/test-utils](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-dom/test-utils/index.d.ts#L280-L307) and [react-test-renderer](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-test-renderer/index.d.ts#L58-L87), neither of which support returning `any` value from their act callbacks.  Even if they technically work now, I don't want our interface (which just returns their utils depending on which renderer is being used) to stray too far from theirs, in case is _does_ become important one day and it breaks our compatibility.

I know it is a little bit clunky, but FWIW, I believe the intended way to use async act ([it's not documented at all](https://reactjs.org/docs/test-utils.html#act) is:

```ts
await act(async () => {
  await hook.result.current.fetchNextPage();
});

andersonvom:
Thanks for the quick feedback, @joshuaellis and @mpeyper!

I see your point, though I wonder if this is legacy behavior as it looks like react itself accepts `T | Thenable<T>` in its [`act` definition][1]. 🤔 

[1]: https://github.com/facebook/react/blob/c7917fe76982c499c81be9571de44fe1293ac893/packages/react/src/ReactAct.js#L17

mpeyper:
Interesting indeed.  I'm not sure where/how that `act` comes into play with everything else, but should the interfaces for the exposed `act` utils get updated to match, it would make some test code bit nicer to read.

PR #777, opened 2022-01-18
BREAKING CHANGE: Update deps

### LINKED ISSUE #729: What's the reason @types are being added as dependencies as opposed to peerDependencies?
https://github.com/testing-library/react-hooks-testing-library/blob/ca13ed9ae8b9c7dc92ffe586e48cbf46f1cfc3a9/package.json#L53-L55

This is causing multiple React declarations errors in our codebase, e.g.

```
node_modules/@testing-library/react-hooks/node_modules/@types/react/index.d.ts:3127:13 - error TS2717: Subsequent property declarations must have the same type.  Property 'datalist' must be of type 'DetailedHTMLProps<HTMLAttributes<HTMLDataListElement>, HTMLDataListElement>', but here has type 'DetailedHTMLProps<HTMLAttributes<HTMLDataListElement>, HTMLDataListElement>'.

3127             datalist: React.DetailedHTMLProps<React.HTMLAttributes<HTMLDataListElement>, HTMLDataListElement>;
                 ~~~~~~~~

  node_modules/@types/react/index.d.ts:3126:13
    3126             datalist: React.DetailedHTMLProps<React.HTMLAttributes<HTMLDataListElement>, HTMLDataListElement>;
                     ~~~~~~~~
    'datalist' was also declared here.
```

joshuaellis:
**What**:

* BREAKING CHANGE: min supported version of node is 12.22
* Updates the deps we have in the package
* Updates the workflow for testing against node versions

**Why**:

* Some upstream deps we want to update require the update in minimum supported node version
* Also _will_ resolve #729

**How**:

**Checklist**:

- [ ] Documentation updated
- [ ] Tests
- [ ] Ready to be merged
      

joshuaellis:
I don't expect this to pass on first go, but wanted to get the ball rolling @mpeyper 

Also on the subject of #729, my POV is move them to devDeps and not include them as peer deps, this is what other libraries typically do from my experience. Happy to be told otherwise though.

joshuaellis:
This is a strange error to have...

mpeyper:
The rules not being found?

I believe `docz` (or one of the related packages) is bringing in an incorrect version of `eslint` which is why we had the version from `kcd-scripts` in our `package.json` to begin with.

joshuaellis:
> I believe docz (or one of the related packages) is bringing in an incorrect version of eslint which is why we had the version from kcd-scripts in our package.json to begin with.

I think you're right. Let me add it to the PR and we'll go from there, as it's (annoyingly) not throwing said error in my env.

mpeyper:
> Also on the subject of #729, my POV is move them to devDeps and not include them as peer deps, this is what other libraries typically do from my experience. Happy to be told otherwise though.

I've just done a quick whip around of a few big react libraries written in TS that I'm aware of and it does appear to be the case that most just list the types as `devDependencies` and nothing else.  I'm not sure if that's because their API does not expose anything from these types (although I think it's unlikely **none** of them expose a `ReactNode` or `ReactElement` somewhere) or if they just accept that there will be a type error when it's not installed.

The only one I could find doing it the way I'm proposing it _should_ be done is [material-ui](https://github.com/mui-org/material-ui/blob/master/packages/mui-material/package.json#L52-L54).  I'm still of the opinion that this is the the _better_ way to handle it as it will give a peer dependency warning if the wrong version is installed, but not complain if the type dependencies are not required (e.g. non-TS users).

I've checked out generated type definitions and only `@types/react` would need to be a `peerDependecy` (and optional in `peerDependenciesMeta`).  `react-dom` and `react-test-renderer` could just move to `devDependencies` as they are not referenced at all in the output.

Issue #796, opened 2022-02-24
ReferenceError: document is not defined in SSR 


jsjoeio:
- `react-hooks-testing-library` version: [v8.0.0-alpha.1](https://github.com/testing-library/react-hooks-testing-library/releases/tag/v8.0.0-alpha.1) - created a commit starting [here](https://github.com/testing-library/react-hooks-testing-library/commit/0543a9e878ae37313408da1f49e7514bd9a55e09)
- `react` version: N/A 
- `react-dom` version (if applicable): N/A
- `react-test-renderer` version (if applicable): N/A
- `node` version: v14.17.6
- `npm` (or `yarn`) version: v1.22.17

### Relevant code or config:

```js
let your = (code, tell) => `the ${story}`
```

### What you did:

I added a test that runs in a Node environment to see if https://github.com/testing-library/react-hooks-testing-library/issues/605 was fixed (sidenote: using this issue/solution in a TypeScript course I'm authoring).

### What happened:

I thought the solution would allow `renderHook` to be called in SSR environments but we get the same issue. 

```typescript
Summary of all failing tests
 FAIL  src/server/__tests__/ssr.test.ts
  ● ssr › should not throw ReferenceError

    The error below may be caused by using the wrong test environment, see https://jestjs.io/docs/en/configuration#testenvironment-string.
    Consider using the "jsdom" test environment.
    
    ReferenceError: document is not defined

      32 |         throw new Error('The component can only be hydrated once')
      33 |       } else {
    > 34 |         container = document.createElement('div')
         |         ^
      35 |         container.innerHTML = serverOutput
      36 |         act(() => {
      37 |           ReactDOM.hydrate(testHarness(renderProps), container!)

      at hydrate (src/server/pure.ts:34:9)
      at Object.<anonymous> (src/server/__tests__/ssr.test.ts:20:5)


Test Suites: 1 failed, 61 passed, 62 total
Tests:       1 failed, 204 passed, 205 total
Snapshots:   0 total
Time:        7.859 s
Ran all test suites.
```

Note: I added the original author on Discord to try and contact them about this PR. I saw in the comments they said they tested in a Node environment so I may be doing something differently.

### Reproduction:

1. `git clone https://github.com/testing-library/react-hooks-testing-library.git`
2. `touch src/server/__tests__/ssr.test.ts`
3. Add this code:

```typescript
/**
 * @jest-environment node
 */
import { useState } from 'react'
import { renderHook, act } from '..'

[...]

PR #799, opened 2022-02-27
feat(server): add SSR test for renderHook

### LINKED ISSUE #605: Usage in close-to-real SSR environment?

jsjoeio:
**What**:

This PR adds a new test based on the work done in #607 to fix #605. Partially related to https://github.com/testing-library/react-hooks-testing-library/issues/796

**Why**:

I noticed a test wasn't added when that bug was fixed. I decided to add a test so we could prevent regressions.

**How**:

A new test was added which runs in a Node environment in Jest. 

**Checklist**:

- [x] Documentation updated
- [x] Tests
- [x] Ready to be merged
      
### DIFF
--- a/src/server/pure.ts
+++ b/src/server/pure.ts
@@ -13,7 +13,7 @@ function createServerRenderer<TProps, TResult>(
 ) {
   let renderProps: TProps | undefined
   let container: HTMLDivElement | undefined
-  let serverOutput: string = ''
+  let serverOutput = ''
   const testHarness = createTestHarness(rendererProps, wrapper, false)
 
   return {
--- a/src/server/pure.ts
+++ b/src/server/pure.ts
@@ -19,13 +19,11 @@ function createServerRenderer<TProps, TResult>(
   return {
     render(props?: TProps) {
       renderProps = props
-      act(() => {
-        try {
-          serverOutput = ReactDOMServer.renderToString(testHarness(props))
-        } catch (e: unknown) {
-          rendererProps.setError(e as Error)
-        }
-      })
+      try {
+        serverOutput = ReactDOMServer.renderToString(testHarness(props))
+      } catch (e: unknown) {
+        rendererProps.setError(e as Error)
+      }
     },
     hydrate() {
       if (container) {
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -145,7 +145,8 @@
       "avatar_url": "https://avatars3.githubusercontent.com/u/3806031?v=4",
       "profile": "https://jsjoe.io",
       "contributions": [
-        "tutorial"
+        "tutorial",
+        "test"
       ]
     },
     {
--- a/README.md
+++ b/README.md
@@ -146,9 +146,9 @@ to test against. It also does not come installed with a specific renderer, we cu
 [`react-test-renderer`](https://www.npmjs.com/package/react-test-renderer) and
 [`react-dom`](https://www.npmjs.com/package/react-dom). You only need to install one of them,
 however, if you do have both installed, we will use `react-test-renderer` as the default. For more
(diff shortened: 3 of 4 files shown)


jsjoeio:
>        Warning: It looks like you called ReactTestUtils.act(...) in a non-browser environment. If you're using TestRenderer for your tests, you should call ReactTestRenderer.act(...) instead of ReactTestUtils.act(...).

I looked at `src/__tests__/autoDetectRenderer.test.ts` because it has a `reactTestRenderer` but it looks custom 🤔 Any tips on how to fix this?

jsjoeio:
Hey @mpeyper - looks like the test is failing. Would love to get your insights to fix this 🙏

mpeyper:
The test looks fine, I’ll see if I can get to the bottom of it.

src/server/pure.ts
```diff
@@ -13,7 +13,7 @@ function createServerRenderer<TProps, TResult>(
 ) {
   let renderProps: TProps | undefined
   let container: HTMLDivElement | undefined
-  let serverOutput: string = ''
+  let serverOutput = ''
   const testHarness = createTestHarness(rendererProps, wrapper, false)
 
   return {
```
mpeyper, line 19: github won't let me comment on the actual line, but the error in the test is being caused by the `act` call on line 22.  For some reason, when you use `react-dom@16.9.0` this tried to use the document which throws an error when testing in the node environment, which doesn't occur in newer versions.

You can replicate this by running:

```sh
npm run install:react-16.9.0
npm test -- --no-watch src/__tests__/ssr.test.ts
```

In the context of server rendering, the `act` call is not really necessary here.  Dropping it out still passes all of our tests and allows your new test to pass with the older version:

```ts
    render(props?: TProps) {
      renderProps = props
      try {
        serverOutput = ReactDOMServer.renderToString(testHarness(props))
      } catch (e: unknown) {
        rendererProps.setError(e as Error)
      }
    },
```

jsjoeio: Wow, I don't think I would have figured that out on my own 😅 I'll test this out and make the necessary changes. Thanks a bunch!

jsjoeio:
Well, that worked! Of course now Netlify is failing lol

```js
12:34:31 PM: error Error in "/opt/build/repo/node_modules/gatsby-plugin-mdx/gatsby-node.js": globalThis is not defined
```

mpeyper:
Yeah, the MetLife build had been failing force while and I’m not sure how to fix it. I’ve mostly been ignoring it with maintenance on this library wrapping up soon anyway.

This all looks fine to me and will try to get it merged next time I’m at my computer.

jsjoeio:
@mpeyper any updates on this by chance? 👀

mpeyper:
Sorry for the delay, and thanks again for the effort.

jsjoeio:
woohoo! thanls for letting me bug you and for the tips!

PR #819, opened 2022-04-10
React 18 cleanup

mpeyper:
**What**:

1. Changing type dependencies to peers deps and restricting  to <18.0
2. Add a note in README and docs about not supporting react 18 in this library
3. ~Update a bunch of dependencies that that have been waiting in other PRs for a while now~
   * Reverted the dependency bumps as it broke both netlify and lint checks

**Why**:

Mostly to stop accidentally resolving type versions that will break things for people and to try to remove any confusion about why we aren't suporting React 18 yet.

**How**:

I've made the type changes a BREAKING CHANGE to be extra cautious for our users in case they need to install the dependencies themselves now, but the update should not require any code changes for anyone.

If all the build check pass for this I will just merge it without waiting for reviews.


mpeyper:
Hmm.... I don't know why netlify is suddenly failing with an error from gatsby.  I guess just the note in README will have to do.

PR #862, opened 2022-06-23
Fix/better error handling for hydrate

amovar18:
**What**:
This PR adds a custom error message for `hydrate` function.

**Why**:
When hydrate function is called in non-browser environment it throws an error `document is not defined` which lead to a non-meaningful error message. This PR now aims to check if document is available or not and then throws a custom error message.

**How**:
This PR just provides a wrapper of an `if else` block inside the `hydrate` function and throws a custom error message saying `Hydrate function can only be called in a client environment with a document available.`.

**Checklist**:

- [ ] Documentation updated
- [x] Tests
- [x] Ready to be merged
      
SideNote: The documentation states that to use `hydrate` function for SSR testing of a hook but the conversation in the issue stated that `render` should be used in non browser environment and `hydrate` should be used on environment where `document` is available. Should the documentation be updated?

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -603,6 +603,15 @@
       "contributions": [
         "doc"
       ]
+    },
+    {
+      "login": "amovar18",
+      "name": "Kudaligi Amoghavarsha",
+      "avatar_url": "https://avatars.githubusercontent.com/u/59614577?v=4",
+      "profile": "https://github.com/amovar18",
+      "contributions": [
+        "maintenance"
+      ]
     }
   ],
   "skipCi": true,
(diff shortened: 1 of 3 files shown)


mpeyper:
Thanks @amovar1. I do not have a heap of time to look at this, but I’ll try to get to it this weekend.

A couple of notes from my quick look now:
1. Can we write a test for this? It should be possible to see the error by calling `hydrate` in [this test file](https://github.com/testing-library/react-hooks-testing-library/blob/main/src/__tests__/ssr.test.ts) as there is no document available there (it’s meant to replicate a ssr environment).
2. I think the docs are ok on this as in the case of  the ssr renderer, `render` simulates the server behaviour, and `hydrate` simulates the client behaviour. That said, updates to add clarity in the docs are always welcome (just be aware that our docs generation is currently broken so might be wasted effort).
3. Speaking of the docs, don’t worry about the netlify errors. This was already broken and unrelated to your changes.

Issue #873, opened 2022-07-15
Tests timeout with jest fakeTimers and waitFor for on Promise.resolve calls


5c077yP:
Hey, I get some of my tests timing out when using waitFor and jest.useFakeTimers, but not using a timer internally, but only `Promise.resolve`. As a context I'm trying to migrate a bigger code base from v4 to the latest version from v5 on some tests are failing. I'm also using jests faketimers by default for the tests.

This might be a bit related to https://github.com/testing-library/react-hooks-testing-library/issues/631

- `react-hooks-testing-library` version: 8.0.1
- `react` version: 17.02
- `react-dom` version (if applicable): 17.02
- `react-test-renderer` version (if applicable): 17.02
- `node` version: 16.14.2
- `yarn` version: 1.22.10

### Relevant code or config:

```js
const React = require('react');
const { renderHook } = require('@testing-library/react-hooks');

jest.useFakeTimers(); // modern or fake doesn't matter here

const useStateAndAsyncEffect = (fn) => {
  const [s, set] = React.useState(false);

  React.useEffect(() => {
    Promise.resolve().then(() => set(true));
  }, []);

  React.useEffect(() => {
    if (s) fn();
  }, [s]);

  return;
};

test('test hook', async () => {
  const fn = jest.fn();
  const { waitFor } = renderHook(() => useStateAndAsyncEffect(fn));
  await waitFor(() => expect(fn).toHaveBeenCalled());
});
```

### What happened:

The test fails from v5 and onwards, but worked in v4.

### Suggested solution:

I was digging a bit into the code and saw v4 is calling `act` inside async-utils inside the `while(true)` loop, while from v5 upwards `act` is only called once. I've tried to figure out the details, but not really sure why calling act more than once is making this work. I'm thinking about react flushing micro tasks more often, but also not very familiar with react internals/fibers.

I've played with patch-package on got this diff working for me. Please let me know what you think about it 🙏 

```diff
diff --git a/node_modules/@testing-library/react-hooks/lib/core/asyncUtils.js b/node_modules/@testing-library/react-hooks/lib/core/asyncUtils.js
index dbf1a49..77b9785 100644
--- a/node_modules/@testing-library/react-hooks/lib/core/asyncUtils.js
+++ b/node_modules/@testing-library/react-hooks/lib/core/asyncUtils.js
@@ -26,9 +26,11 @@ function asyncUtils(act, addResolver) {
 
     const waitForResult = async () => {
       while (true) {
[...]

PR #878, opened 2022-07-19
feat(waitFor): Automatically advance Jest fake timers

eps1lon:
**What**:

Closes https://github.com/testing-library/react-hooks-testing-library/issues/873

**Why**:

There's little value in `waitFor` if one has to advance timers by themselves. 

**How**:

Copy implementation from `@testing-library/dom`. 

There's an opportunity to write a more generic `waitFor` that we can share across Testing Library that specific host configs build on top on (e.g. also running MutationObserver in `/dom`). But that requires more coordination (starting with an extensive test suite). For now this hopefully suffices (pending testing in actual codebases).

**Checklist**:

- [ ] Documentation updated
- [x] Tests
- [ ] Test in actual codebase
- [ ] Ready to be merged
      

src/core/asyncUtils.ts
```diff
@@ -42,6 +63,107 @@ function asyncUtils(act: Act, addResolver: (callback: () => void) => void): Asyn
...
+
+        if (finished) {
+          break
+        }
+
+        // In this rare case, we *need* to wait for in-flight promises
+        // to resolve before continuing. We don't need to take advantage
+        // of parallelization so we're fine.
+        // https://stackoverflow.com/a/59243586/971592
+        await act(async () => {
```
eps1lon, line 122: This is useless for React 17 and should be removed to not confuse. It is only needed in 18 where we need to remove the `act` from `waitFor`

PR #924, opened 2022-11-18
chore(docs): add waitForValueToChange shim to migration guide

develohpanda:
**What**: Add tips to migration guide

**Why**: Suggestion from comment in https://github.com/testing-library/react-testing-library/pull/991#issuecomment-1319459830

**Checklist**:

- [X] Documentation updated
- [ ] Tests
- [X] Ready to be merged
      
### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -115,6 +115,20 @@ APIs. A similar result can be achieved by using `waitFor`:
 +});
 ```
 
+Alternatively this utility can be implemented as the following, for a direct migration:
+
+```ts
+import { waitFor } from '@testing-library/react';
+
+const waitForValueToChange = async <T,>(getValue: () => T) => {
+    const original = getValue();
+
+    await waitFor(async () => {
+        expect(await original).not.toBe(await getValue());
+    });
+};
+```
+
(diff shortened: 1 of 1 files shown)


MIGRATION_GUIDE.md
```diff
@@ -115,6 +115,20 @@ APIs. A similar result can be achieved by using `waitFor`:
 +});
 ```
 
+Alternatively this utility can be implemented as the following, for a direct migration:
+
+```ts
+import { waitFor } from '@testing-library/react';
+
+const waitForValueToChange = async <T,>(getValue: () => T) => {
```
mpeyper, line 123: **nitpick:** I don't think the "," in `<T,>` is doing anything, right?

mpeyper (approved):
I'll wait to see how you feel about the comment above, but other than that I'm happy to merge this.

MIGRATION_GUIDE.md
develohpanda: Apologies, you're right! I had this written in a `.tsx` file where the `,` was necessary to differentiate from a jsx tag.

```suggestion
const waitForValueToChange = async <T>(getValue: () => T) => {
```

